PR #2, opened 2015-10-19
Composing structs

nrc:
@nikomatsakis, do you have any suggestions for improving this description? Was this the pattern you had in mind?

### DIFF
--- a/patterns/compose-structs.md
+++ b/patterns/compose-structs.md
@@ -1,0 +1,101 @@
+# Compose structs together for better borrowing
+
+TODO - this is not a very snappy name
+
+## Description
+
+Sometimes a large struct will cause issues with the borrow checker - although
+fields can be borrowed independently, sometimes the whole struct ends up being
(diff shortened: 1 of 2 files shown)


nikomatsakis:
This is roughly what I had in mind, yes. My main suggestion for improvement would be making a more realistic example -- I can go cast about to find something that's actually come up for me.

nikomatsakis:
@nrc sorry for the delay, forgot about this :(

nrc:
Cool, I'll merge then. Would be great to have a PR with a better example.

patterns/compose-structs.md
```diff
@@ -0,0 +1,101 @@
+# Compose structs together for better borrowing
```
posborne: Maybe "Decompose struct for better borrowing" would be better.  The act, if one were refactoring, here seems to be one of decomposition rather than composition.  Saying "Compose ... together" is redundant.

PR #13, opened 2015-10-27
fix compile error in comment

liigo:


idioms/concat-format.md
```diff
@@ -13,7 +13,7 @@ non-literal strings.
 ```rust
 fn say_hello(name: &str) -> String {
     // We could construct the result string manually.
-    // let mut result = "Hello";
+    // let mut result = "Hello".to_string();
```
nrc: `to_owned` is preferable to `to_string`

liigo:
@nrc comment addressed

nrc:
Thanks!

PR #23, opened 2016-06-02
Adding 'clone to satisfy the borrow checker' anti-pattern

liamzdenek:
I feel like this write-up could be improved, but I'm not exactly sure how -- recommendations and modifications are welcome.

### DIFF
--- a/anti_patterns/borrow_clone.md
+++ b/anti_patterns/borrow_clone.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-The borrow checker prevents rust users from developing otherwise unsafe code by
+The borrow checker prevents Rust users from developing otherwise unsafe code by
 ensuring that either: only one mutable reference exists, or potentially many but
 all immutable references exist. If the code written does not hold true to these 
 conditions, this anti-pattern arises when the developer resolves the compiler
@@ -11,22 +11,21 @@ error by cloning the variable.
 
 ## Example
 
-```rust fn main() {
-    // define any variable
-    let mut x = 5;
-    
-    // Borrow `x` -- but clone it first
-    let y = &mut (x.clone()); 
+```rust
+// define any variable
+let mut x = 5;
 
-    // perform some action on the borrow to prevent rust from optimizing this
-    //out of existence
-    *y += 1;
+// Borrow `x` -- but clone it first
+let y = &mut (x.clone()); 
 
-    // without the x.clone() two lines prior, this line would fail on compile as
-    // x has been borrowed
-    // thanks to x.clone(), x was never borrowed, and this line will run.
-    println!("{}", x);
-}
+// perform some action on the borrow to prevent rust from optimizing this
+//out of existence
+*y += 1;
+
+// without the x.clone() two lines prior, this line would fail on compile as
+// x has been borrowed
+// thanks to x.clone(), x was never borrowed, and this line will run.
+println!("{}", x);
 ```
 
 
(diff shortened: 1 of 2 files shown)


anti_patterns/borrow_clone.md
```diff
@@ -0,0 +1,56 @@
...
+The borrow checker prevents rust users from developing otherwise unsafe code by
+ensuring that either: only one mutable reference exists, or potentially many but
+all immutable references exist. If the code written does not hold true to these 
+conditions, this anti-pattern arises when the developer resolves the compiler
+error by cloning the variable.
+
+
+## Example
+
+```rust fn main() {
```
lfairy: The "rust" and "fn main()" should be on separate lines, like this:

    ```rust
    fn main() {

anti_patterns/borrow_clone.md
```diff
@@ -0,0 +1,56 @@
+# Clone to satisfy the borrow checker
+
+## Description
+
+The borrow checker prevents rust users from developing otherwise unsafe code by
```
lfairy: "Rust" should be capitalized

anti_patterns/borrow_clone.md
lfairy: Also, I don't think you need to wrap the code in `fn main()` -- the other examples in this repo (along with the official docs) keep the code bare.

anti_patterns/borrow_clone.md
```diff
@@ -0,0 +1,55 @@
...
+
+// perform some action on the borrow to prevent rust from optimizing this
+//out of existence
+*y += 1;
+
+// without the x.clone() two lines prior, this line would fail on compile as
+// x has been borrowed
+// thanks to x.clone(), x was never borrowed, and this line will run.
+println!("{}", x);
+```
```
cbreeden: I think we can come up with a better example; I'm not sure if this is really a problem when your type is copy.  Let me play around with some examples and get back to you.

cbreeden:
I like your motivation and description for the most part.  I think we can make improve this section quite a bit if we put a little more polish into the example, and add examples that shows how you can avoid the problems.  I'll play around with a few ideas and get back to you.

cbreeden:
So I tried to think of some good examples, what do you think about this one

```
    let mut vec = vec![("Bob".to_string(), 17), ("Stacy".to_string(), 21)];
    let new_pair = ("Ferris".to_string(), 13);
    
    // check for duplicates
    let (ref key, _) = new_pair;
    if vec.iter().map(|&(ref k, _)| k).find(|&k| k == key).is_none() {
        // Can't move since new_pair is being referenced still!
        vec.push(new_pair);
    }
    
    println!("{:?}", vec);
```

I could imagine someone trying to add a `.clone()` to the `new_pair` in `vec.push(new_pair);`.  A lot of times these problems can be solved by properly managing their lifetimes -- in this case just adding `k == &new_pair.0`.  In other cases, closing a mutable borrow scope in brackets.  I don't know if this is the best example, just throwing it out there.

llogiq:
I have a few ideas for examples:

* change an enum value, leaving parts intact (see #31)
* select & remove items from a collection (use Drain instead once available)
* un-by-ref an iterator (`.cloned()`) which is actually OK in many cases (unless you can consume the collection while iterating, which should be preferred)
* `entry(_.clone())` – there's an RFC to get rid of that.

Hope that helps.

pickfire:
Anything left for this?

simonsan:
@liamzdenek any updates on this?

simonsan:
Closing due to inactivity. Further additions to the PR can be made in #110.

PR #25, opened 2016-07-25
Add "Treating Option like a list"

lfairy:
Not sure it's okay to reference a crates.io library, but the article would have been really short otherwise

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ language.
 * TODO leak amplification ("Vec::drain sets the Vec's len to 0 prematurely so that mem::forgetting Drain "only" mem::forgets more stuff. instead of exposing uninitialized memory or having to update the len on every iteration")
 * [Finalisation in destructors](idioms/dtor-finally.md)
 * TODO interior mutability - UnsafeCell, Cell, RefCell
-* TODO treating Option like a list
+* [Iterating over an `Option`](idioms/option-iter.md)
 * TODO `Default` trait
 
 ### Design patterns
--- a/idioms/option-iter.md
+++ b/idioms/option-iter.md
@@ -1,0 +1,46 @@
+# Iterating over an `Option`
+
+## Description
+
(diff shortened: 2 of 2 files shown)


idioms/option-list.md
```diff
@@ -0,0 +1,29 @@
+# Treating `Option` like a list
```
nrc: Perhaps a better framing of this idiom is "Iterating an Option", you could then reference ref_slice as an added extra.

idioms/option-list.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Examples
+
+```rust
+let arg = Some("doovalacky");
+// equivalent to `if let Some(value) = ...`
+for value in arg {
+    println!("Received: {}", value);
+}
+```
```
nrc: A better example might be using `Chain` to append an Option to another iterator, and/or a function that takes `T: Iterator<U>`

I would sat that using a `for` loop with an `Option` is less idiomatic than using `if let` (and it might be worth stating that).

idioms/option-list.md
```diff
@@ -0,0 +1,29 @@
+# Treating `Option` like a list
+
+## Description
+
+`Option` can be viewed as a list that contains either zero or one elements. It can be iterated over using `Option::iter()`, and converted to a slice using the `ref_slice` crate. Keeping these tools in mind can lead to simpler, more idiomatic code.
```
nrc: and work with generic code.

nrc:
@lfairy very sorry for letting this PR sit for so long - I put it off due to being busy and then it slipped my mind.

Thanks for submitting the PR and I hope to be much quicker with feedback in the future.

cc @cbreeden (if you have time to comment)

lfairy:
@nrc thanks for the feedback! I've pushed a new version that should fix the issues you brought up.

re-r?

nrc:
looks great, thank you!

PR #28, opened 2016-09-23
Clone to closure idiom

kriomant:


idioms/clone-to-closure.md
```diff
@@ -0,0 +1,33 @@
...
+to move environment. However, often you want to give copy of some data to closure.
+Use variable rebinding in separate scope for that.
+
+
+## Example
+
+```rust
+let mut num = Rc::new(5);
+let closure = {
+	let num = num.clone();
```
lfairy: This code should use four-space indents, not tabs, to stay consistent with other articles.

idioms/clone-to-closure.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+
+## Advantages
+
+Copied data are grouped together with closure definition, so their purpose is more clear
+and they will be dropped immediately even if they are not consumed by closure.
+
+Closure uses same variable names as surrounding code whether data are copied or moved.
+
```
lfairy: As a disadvantage, you should mention rightward drift: adding another scope will indent the code further, which can make it harder to read.

idioms/clone-to-closure.md
kriomant: fixed

idioms/clone-to-closure.md
kriomant: done

nrc:
I think the name of the pattern is a little misleading - it applies just as much to by reference variables captured by a `move` closure as it does to cloned ones.

idioms/pass-var-to-closure.md
```diff
@@ -1,10 +1,10 @@
-# Pass clone to closure
+# Pass variable to closure
```
lfairy: This should be "variables"

lfairy:
Looks good to me. Since we've had two people review it already, I'll go ahead and merge.

Thanks @kriomant for your contribution!

Issue #27, opened 2016-09-23
Using `&str` vs using `&String`


cbreeden:
This is a common idiom, where using an `&str` is often more flexible than using a `&String`.

MarcoIeni:
Same is true for Path and PathBuf I believe, and also for Vec and slices.
There is also a clippy lint about this (works only with nightly): https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
We should express this concept in a general way and add it to idioms.

simonsan:
This issue should be fixed with #29 I believe, if not feel free to reopen.

PR #29, opened 2016-09-23
Use coercion for arguments

cbreeden:
Describe some of the reasons it is considered idiomatic to use `&str` over `&String` in most cases.

ref.: #27


idioms/str-vs-string.md
```diff
@@ -0,0 +1,90 @@
...
+fn three_vowles(word: &str) -> bool {
+```
+
+then the previous example will not print
+
+``` bash
+Ferris: false
+```
+
+But wait, that's not all!  There is more to this story.  It's very likely that you may say: that doesn't matter, I will never be using a `&'static str` as an input anways (as we did when we used `"Ferris"` as an input to our example).  Even ignoring this case we will find that using `&str` will give us more flexibility than using `&String`.  Let's now take an example where someone gives us a sentance, and we want to determine if any of the words in the sentence has a word that contains three consecutive vowels.  We probably should make use of the function we have already defined and simply feed in each word from the sentence.  An example of this could look like this:
```
FraGag: s/sentance/sentence/

idioms/str-vs-string.md
```diff
@@ -0,0 +1,90 @@
...
+``` bash
+curious has three consecutive vowels!
+```
+
+However, this example will not run when our function is declared with an argument type of `&String`.  This is because string slices will yield a `&str` and not a `&String`.  One way you can thing about this is: to have a `&String` you will need a reference to a `String` object which requires three things: pointer to your data, then length of your data, and how much capacity you have left to insert new data.  This last entry doesn't make much sense when you are talking about string slices as string slices only care about a window of your data needing only: a pointer to your data, and the length of your data slice.
+
+## See also
+
+TODO: Add references to the book, rustbyexample, and possibly steve's post on the differences between a `String` and `str`.
+TODO: Possibly reorginize to have a smaller example first and this larger example under a `## Discussion` section.
```
FraGag: s/reorginize/reorganize/

idioms/str-vs-string.md
```diff
@@ -0,0 +1,90 @@
...
+## Description
+
+It is usually recommended to prefer a `&str` type over a `&String` type.  In most cases `&str` has more flexibility.  An `&String` will coerce to a `&str` type when required -- however the oppositive is not true.  
+
+## Example
+
+Let's look at a few examples.  Consider an example where we wish to determine if a word contains three consecutive vowels.  We certainly don't need to own the string to determine this so we will take a reference.  The code might look something like this:
+
+``` rust
+fn has_three_vowles(word: &String) -> bool {
```
FraGag: "vowles"? Shouldn't this function be named either `has_three_vowels`?

idioms/str-vs-string.md
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+This example will work fine, as shown here:
+
+``` rust
+fn main() {
+    let ferris = "Ferris".to_string();
+    let curious = "Curious".to_string();
+    println!("{}: {}", ferris, three_vowles(&ferris));
```
FraGag: The function is defined as `has_three_vowles`, but all references use the name `three_vowles`.

idioms/str-vs-string.md
cbreeden: thanks for catching all of these for me!

cbreeden:
@lfairy Does this look fine to you?

idioms/str-vs-string.md
```diff
@@ -0,0 +1,89 @@
+# Using &str instead of &String
+
+## Description
+
+It is usually recommended to prefer a `&str` type over a `&String` type for arguments.  In most cases `&str` has more flexibility.  An `&String` will coerce to a `&str` type when required -- however, the oppositive is not true.
```
FraGag: s/oppositive/opposite/

lfairy: "An `&String`" should be "A `&String`"

Also, did you mean "opposite" or "converse"? "Oppositive" is a real word, but it's rare.

cbreeden: From here on out, I will claim that auto-correct is responsible for every typo I make :) thanks!

idioms/str-vs-string.md
```diff
@@ -0,0 +1,89 @@
...
+}
+```
+
+Running this example using our function declared with an argument type `&str` will yield
+
+``` bash
+curious has three consecutive vowels!
+```
+
+However, this example will not run when our function is declared with an argument type `&String`.  This is because string slices are a `&str` and not a `&String`.  One way you can think about why this is the case: to have a `&String` you will need a reference to a `String` object which requires three things: a pointer to your data, the length of your data, and how much capacity you have left to insert new data.  This last entry doesn't make much sense when you are talking about string slices. A string slices only cares about a window of your data, needing only: a pointer to your data, and the length of your data slice.
```
lfairy: It feels like this explanation should belong in the introduction, not tucked away at the end after all the examples.

idioms/str-vs-string.md
```diff
@@ -0,0 +1,89 @@
...
+It is usually recommended to prefer a `&str` type over a `&String` type for arguments.  In most cases `&str` has more flexibility.  A `&String` will coerce to a `&str` type when required -- however, the opposite is not true.  
+
+## Example
+
+Let's look at a few examples.  Consider an example where we wish to determine if a word contains three consecutive vowels.  We certainly don't need to own the string to determine this so we will take a reference.  The code might look something like this:
+
+``` rust
+fn three_vowels(word: &String) -> bool {
+    let mut vowel_count = 0;
+    for char in word.chars() {
```
lfairy: `char` is the name of a built-in type, so you should use another name (e.g. `c`) to avoid confusion.

idioms/str-vs-string.md
```diff
@@ -0,0 +1,89 @@
...
+fn three_vowels(word: &str) -> bool {
+```
+
+then the previous example will now print
+
+``` bash
+Ferris: false
+```
+
+But wait, that's not all!  There is more to this story.  It's likely that you may say to yourself: that doesn't matter, I will never be using a `&'static str` as an input anways (as we did when we used `"Ferris"`).  Even ignoring this special example, you may still find that using `&str` will give you more flexibility than using a `&String`.  Let's now take an example where someone gives us a sentence, and we want to determine if any of the words in the sentence has a word that contains three consecutive vowels.  We probably should make use of the function we have already defined and simply feed in each word from the sentence.  An example of this could look like this:
```
lfairy: > But wait, that's not all!

I like the informal tone throughout this article! This in particular is really good.

idioms/str-vs-string.md
cbreeden: Good idea

cbreeden:
> I wonder if this article should be about borrowed types in general (e.g. talk about Vec<T> and &[T] as well)

This is a good point.  Let me think about that.

idioms/str-vs-string.md
cbreeden: whoops

idioms/str-vs-string.md
```diff
@@ -0,0 +1,89 @@
+# Using &str instead of &String
+
+## Description
+
+It is usually recommended to prefer a `&str` type over a `&String` type for arguments.  In most cases `&str` has more flexibility.  A `&String` will coerce to a `&str` type when required -- however, the opposite is not true.
```
nrc: I would be stronger here - there is never any reason to use `&String` except in generic code. There is sometimes reason to use `&mut String`, but that is rare.

nrc: Also, `&String` is less efficient than `&str` because it has two levels of indirection rather than one.

nrc: And also again, it is possible (though I'm not aware of anyone actually doing this) for alternate `String` implementations to also implement `Deref<Target=str>` and thus taking `&str` accommodates this use case too.

cbreeden: Thanks for the feedback.  I wasn't comfortable to be more assertive there (though I did think about it), and I mentioned about about the size differences towards the end but not the additional layer of indirection, I'll add that as well.

In fact, @lfairy suggested that I generalize the entire article to include things like `&Vec<T>` vs `&[T]` and I wasn't sure what the proper generalization would be.  But now I think the proper generalization is: for slice-able fat pointer types.

nrc: I don't think you need to limit to slice-able or fat pointer types. You should always prefer using the borrowed type over borrowing the owned type, e.g., `&T` to `&Box<T>`

cbreeden:
@nrc @lfairy Do you like this approach better?

idioms/coercion-arguments.md
```diff
@@ -1,18 +1,18 @@
-# Using &str instead of &String
+# Using coercion for arguments
```
lfairy: I think it's better to emphasize the idea of using *borrowed types*, rather than the deref operations / coercions that target them.

It was always more idiomatic to use the borrowed types, even before we had deref coercions as a language feature (i.e. rust-lang/rfcs#241). The coercions just made them easier to use.

idioms/coercion-arguments.md
```diff
@@ -1,18 +1,18 @@
...
 
 ## Description
 
-It is usually recommended to prefer a `&str` type over a `&String` type for arguments.  In most cases `&str` has more flexibility.  A `&String` will coerce to a `&str` type when required -- however, the opposite is not true.  
+Usings a target of a deref coericion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.  For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`.  Not only that, you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as can be illustrated in each of the previous examples.  For instance, a `String` has a layer of indirection, so a `&String` will have two layers of indrection.  We can avoid this by using `&str` instead, and letting `&String` coerce to a `&str` whenever the function is invoked.  And finally, using deref targets for a function can increase the number of viable inputs for your function which may increase its utility.
```
lfairy: Nit: "using" not "usings"

lfairy: I think the first and last sentences of this paragraph are saying the same thing? Feels like this needs a refactor -- I'd put the "more flexibility" and "fewer layers of indirection" points in separate paragraphs, at least.

idioms/coercion-arguments.md
```diff
@@ -56,10 +57,12 @@ This example fails because a `&str` type will not coerce to a `&String` type.  W
 fn three_vowels(word: &str) -> bool {
 ```
 
-then the previous example will now print
+then the both of the previous examples will compile and print the same output.
```
lfairy: Nit: extra "the"

idioms/coercion-arguments.md
```diff
@@ -82,7 +85,7 @@ Running this example using our function declared with an argument type `&str` wi
 curious has three consecutive vowels!
 ```
 
-However, this example will not run when our function is declared with an argument type `&String`.  This is because string slices are a `&str` and not a `&String`.  One way you can think about why this is the case: to have a `&String` you will need a reference to a `String` object which requires three things: a pointer to your data, the length of your data, and how much capacity you have left to insert new data.  This last entry doesn't make much sense when you are talking about string slices. A string slices only cares about a window of your data, needing only: a pointer to your data, and the length of your data slice.
+However, this example will not run when our function is declared with an argument type `&String`.  This is because string slices are a `&str` and not a `&String`.
```
lfairy: Should mention how `&str` can't be coerced to `&String`, because the other direction does work.

e.g. "This is because string slices are of type `&str`, which cannot be converted to `&String`"

nrc: Well, technically you can convert but it requires an allocation and is not implicit, whereas String -> str is cheap and implicit

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,92 @@
+# Using coercion for arguments
```
pickfire, line 1:
```suggestion
# Use coercion for arguments
```
I think `Use` is better than `Using`, title should also be updated in index.

simonsan:
@lambda-fairy Worked in some fixes from you, would you mind reading over it one last time? ;-)

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,108 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument which may increase its utility as it increases the number of viable inputs for this function.
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`. 
+Not only that, you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as can be illustrated in each of the previous examples.
```
simonsan, line 7:
```suggestion
Not only that, you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as it will be illustrated in the following examples. 
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,112 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument which may increase its utility as it increases the number of viable inputs for this function.
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`.
```
MarcoIeni, line 6:
```suggestion
For example; using `&str` instead of `&String`, or `&Path` instead of `&PathBuf` or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`. 
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,112 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument which may increase its utility as it increases the number of viable inputs for this function.
```
MarcoIeni, line 5:
```suggestion
Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.
In this way, the function will accept more input types.
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,112 @@
...
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`. 
+Not only that, you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as it will be illustrated in the following examples. 
+For instance, a `String` has a layer of indirection, so a `&String` will have two layers of indrection.
+We can avoid this by using `&str` instead, and letting `&String` coerce to a `&str` whenever the function is invoked. 
+
+## Example
+
+For this example, we will illustrate some differences for using `&String` as a function argument versus using a `&str`, but the ideas easily apply to using `&Vec<T>` versus using a `&[T]`, or even using a `&T` versus a `&Box<T>`.
```
MarcoIeni: This is a repetition of line 6. Let's keep this or the other one.

simonsan: updated it already, press "refresh" :-P

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,112 @@
...
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`, or even `&T` as opposed to a `&Box<T>`. 
+Not only that, you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as it will be illustrated in the following examples. 
+For instance, a `String` has a layer of indirection, so a `&String` will have two layers of indrection.
+We can avoid this by using `&str` instead, and letting `&String` coerce to a `&str` whenever the function is invoked. 
+
+## Example
+
+For this example, we will illustrate some differences for using `&String` as a function argument versus using a `&str`, but the ideas easily apply to using `&Vec<T>` versus using a `&[T]`, or even using a `&T` versus a `&Box<T>`. 
+Consider an example where we wish to determine if a word contains three consecutive vowels.
+We don't need to own the string to determine this so we will take a reference.
```
MarcoIeni, line 18:
```suggestion
We don't need to own the string to determine this, so we will take a reference.
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,115 @@
...
+    let curious = "Curious".to_string();
+    println!("{}: {}", ferris, three_vowels(&ferris));
+    println!("{}: {}", curious, three_vowels(&curious));
+
+    // Using a `&String` type in our arguent we will find the following example fails:
+    // println!("Ferris: {}", three_vowels("Ferris"));
+    // println!("Curious: {}", three_vowels("Curious"));
+
+}
+```
```
MarcoIeni, line 50:
```suggestion
}
TODO REMOVE ME```

This works fine because we are passing a `&String` type as a parameter.

What if we try to pass `&str` instead?

TODO REMOVE ME```rust,ignore
println!("Ferris: {}", three_vowels("Ferris"));
println!("Curious: {}", three_vowels("Curious"));
TODO REMOVE ME```
```

MarcoIeni: I had to insert TODO REMOVE ME because if you want to do github suggestions you can't mess up with code blocks

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,115 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.
+In this way, the function will accept more input types.
+
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`.  This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__, e.g., `&T` to `&Box<T>`.
```
MarcoIeni, line 8:
```suggestion
For example; using `&str` instead of a `&String`, or `&Path` instead of `&PathBuf`, or `&[T]` in preference of `&Vec<T>`.  This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__, e.g., `&T` to `&Box<T>`.
```

MarcoIeni:
Other than the 2 open suggestions I've made it looks good to me.

idioms/coercion-arguments.md
simonsan: I've put that inside the function before and commented it out to let people try it in the playground.

simonsan: Pushed the last version I wrote now, can you read it if it's fine?

idioms/coercion-arguments.md
simonsan: I'm not sure if we should state that without discussing the general differences of `Path` and `PathBuf`, I feel the example without stating it is much clearer for the comparisons.

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
...
+
+Running this example using our function declared with an argument type `&str` will yield
+
+```bash
+curious has three consecutive vowels!
+```
+
+However, this example will not run when our function is declared with an argument type `&String`. 
+This is because string slices are a `&str` and not a `&String` which would require an allocation to be converted to `&String` which is not implicit, whereas converting from `String` to `&str` is cheap and implicit.
+
```
simonsan, line 114: Missing:
- Discussion
- Advantages
- Disadvantages

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
...
+    false
+}
+
+fn main() {
+    let ferris = "Ferris".to_string();
+    let curious = "Curious".to_string();
+    println!("{}: {}", ferris, three_vowels(&ferris));
+    println!("{}: {}", curious, three_vowels(&curious));
+
+    // Using a `&String` type in our argument we will find the following example fails:
```
MarcoIeni, line 45:
```suggestion
    // This works fine, but the following two lines would fail:
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.
+In this way, the function will accept more input types.
+
+For example; using `&str` instead of a `&String`, or `&[T]` in preference of `&Vec<T>`.  This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__, e.g., `&T` to `&Box<T>`.
```
pickfire, line 8:
```suggestion
This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__. E.g., `&str` over `&String`, `&[T]` over `&Vec<T>` or `&T` over `&Box<T>`.
```

Looks duplicated to have examples twice in both places. We should make it one, I saw the previous comment but it was resolved.

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
...
+            'a' | 'e' | 'i' | 'o' | 'u' => {
+                vowel_count += 1;
+                if vowel_count >= 3 {
+                    return true
+                }
+            }
+            _ => vowel_count = 0
+        }
+    }
+    false
```
pickfire, line 36: Would using bytes here be a better choice?

MarcoIeni: bytes instead of characters? Why? It might look scary for beginners.

idioms/coercion-arguments.md
MarcoIeni: mmm..Do you mean that this chapter should be restructured to include those?

Because I don't feel like other content should be added. There is the risk that we create repetitions.

idioms/coercion-arguments.md
pickfire: Is it really that scary, it's just low-effort change from `'a'` to `b'a'` and `chars()` to `bytes()` for a lot more performance?

idioms/coercion-arguments.md
simonsan: I felt that this could take out repetition and bundle it under the mentioned points instead of spreading it over the article. 🤔 Just an idea though.

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
...
+```bash
+curious has three consecutive vowels!
+```
+
+However, this example will not run when our function is declared with an argument type `&String`. 
+This is because string slices are a `&str` and not a `&String` which would require an allocation to be converted to `&String` which is not implicit, whereas converting from `String` to `&str` is cheap and implicit.
+
+## See also
+
+For more discussion on how to handle `String` and `&str` see [this blog series (2015)](https://web.archive.org/web/20201112023149/https://hermanradtke.com/2015/05/03/string-vs-str-in-rust-functions.html) by Herman J. Radtke III.
```
simonsan, line 117:
```suggestion
## See also
- [Rust Language Reference on Type Coercions](https://doc.rust-lang.org/reference/type-coercions.html)
- For more discussion on how to handle `String` and `&str` see [this blog series (2015)](https://web.archive.org/web/20201112023149/https://hermanradtke.com/2015/05/03/string-vs-str-in-rust-functions.html) by Herman J. Radtke III.
```

idioms/coercion-arguments.md
simonsan: That would be good if you would do that in your own code when using a `three_vowels` function that is optimized for performance, but I think for examples it's more about what they show overall, and this example is not about performance or `chars vs bytes` but about `&str vs &String`  or Coercion in function arguments. So as I know what you mean, I think changing this to bytes is not overly relevant to this example.

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.
+In this way, the function will accept more input types.
+
+This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__. E.g., `&str` over `&String`, `&[T]` over `&Vec<T>` or `&T` over `&Box<T>`.
+
+Using borrowed types you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection, as it will be illustrated in the following examples. For instance, a `String` has a layer of indirection, so a `&String` will have two layers of indrection.
```
lambda-fairy, line 10: "as it will be illustrated in the following examples" is redundant and can be removed.

```suggestion
Using borrowed types you can avoid layers of indirection for those instances where the owned type already provides a layer of indirection. For instance, a `String` has a layer of indirection, so a `&String` will have two layers of indrection.
```

idioms/coercion-arguments.md
```diff
@@ -0,0 +1,117 @@
+# Use borrowed types for arguments
+
+## Description
+
+Using a target of a deref coercion can increase the flexibility of your code when you are deciding which argument type to use for a function argument.
+In this way, the function will accept more input types.
+
+This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__. E.g., `&str` over `&String`, `&[T]` over `&Vec<T>` or `&T` over `&Box<T>`.
```
lambda-fairy, line 8: Nit: I prefer Oxford commas, especially in complex sentences like this.

```suggestion
This is not limited to slice-able or fat pointer types. In fact you should always prefer using the __borrowed type__ over __borrowing the owned type__. E.g., `&str` over `&String`, `&[T]` over `&Vec<T>`, or `&T` over `&Box<T>`.
```

idioms/coercion-arguments.md
lambda-fairy: Agreed. Using bytes introduces another concept (UTF-8 encoding) which distracts from the main point.

idioms/coercion-arguments.md
lambda-fairy: I'm happy to land this now and iterate on it later. This document might not be perfect but I think it's good enough, and a rewrite will probably drag this PR out too much.

lambda-fairy (approved):
LGTM modulo comments

idioms/coercion-arguments.md
simonsan: Agreed ;-)

simonsan (approved):

simonsan:
> LGTM modulo comments

That was a ride 🎢 🙄 

:-D

simonsan:
Thanks for the PR @cbreeden ;-)

PR #30, opened 2016-10-05
Prefer small crates

lfairy:


patterns/small-crates.md
```diff
@@ -0,0 +1,31 @@
...
+* Small crates are easier to understand and audit.
+* Since the compilation unit of Rust is the crate, splitting a project into multiple crates can allow more of the code to be built in parallel.
+* If published on crates.io, a crate can be reused by other projects.
+
+## Disadvantages
+
+* Too many crates can lead to code that is hard to follow.
+* This can lead to "dependency hell", where a project depends on multiple conflicting versions of a crate at the same time.
+* Packages on crates.io are not curated. A crate may be poorly written, have unhelpful documentation, or be outright malicious.
+
```
nrc: Two small crates will be less optimised than one large one, since we don't do LTO by default (and even if you do it, it is not as good as intra-unit optimisation).

patterns/small-crates.md
```diff
@@ -0,0 +1,31 @@
...
+## Advantages
+
+* Small crates are easier to understand and audit.
+* Since the compilation unit of Rust is the crate, splitting a project into multiple crates can allow more of the code to be built in parallel.
+* If published on crates.io, a crate can be reused by other projects.
+
+## Disadvantages
+
+* Too many crates can lead to code that is hard to follow.
+* This can lead to "dependency hell", where a project depends on multiple conflicting versions of a crate at the same time.
```
nrc: Is this actually a problem? Other than build times/binary size?

patterns/small-crates.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Advantages
+
+* Small crates are easier to understand and audit.
+* Since the compilation unit of Rust is the crate, splitting a project into multiple crates can allow more of the code to be built in parallel.
+* If published on crates.io, a crate can be reused by other projects.
+
+## Disadvantages
+
+* Too many crates can lead to code that is hard to follow.
```
nrc: I feel like this needs explaining - in theory, it should not be any harder to follow code into a separate crate than a separate module.

patterns/small-crates.md
```diff
@@ -0,0 +1,31 @@
+# Prefer small crates
+
+## Description
+
+Prefer small crates that do one thing well.
+
+Cargo and crates.io make it easy to add third-party libraries, much more so than in say C or C++. We should take advantage of this tooling, and use smaller, more fine-grained dependencies.
+
```
nrc: Might be worth mentioning why the 'right pad' issue does not affect Rust

patterns/small-crates.md
lfairy: I've seen the issue with Piston, e.g. PistonDevelopers/Piston-Tutorials#76. If two crates depend on different versions of the same trait, then we might get "trait not implemented" errors when using them together.

That kind of problem is pretty rare though, and it's easy to mitigate (re-export dependent crates). I wonder if "re-export dependencies" should be another pattern.

softprops: In the past, I've had similar issues with an error message like ... expected a 'foo::Bar' got a 'foo::Bar'.

lfairy: @softprops Yeah, that's the kind of issue I was talking about.

nrc:
Looks good, thanks for the changes!

PR #31, opened 2016-10-08
new idiom: mem::replace

llogiq:
This is something I recently found to be quite useful.

### DIFF
--- a/idioms/mem-replace.md
+++ b/idioms/mem-replace.md
@@ -1,0 +1,83 @@
+# `mem::replace` to keep owned values in changed enums
+
+## Description
+
+Say we have a `&mut MyEnum` which has (at least) two variants,
+`A { name: String, x: u8 }` and `B { name: String }`. Now we want to change
+`MyEnum::A` to a `B` if `x` is zero, while keeping `MyEnum::B` intact.
+
+We can do this without cloning the `name`.
+
+## Example
+
+```rust
+use std::mem;
+
+fn a_to_b(e: &mut MyEnum) {
+    // we mutably borrow `e` here. This precludes us from changing it directly
+    // as in `*e = ...`, because the borrow checker won't allow it. Of course
+    // we could just take a reference to `name` and clone that, but why pay an
+    // extra allocation for something we already have?
+    let have_name = match *e {
+        MyEnum::A(ref mut name, x) if x == 0 => {
+            // this takes out our `name` and put in an empty String instead
+            // note that empty strings don't allocate
+            Some(mem::replace(name, "".to_string()))
+        }
+        // nothing to do in all other cases
+        _ => None
+    };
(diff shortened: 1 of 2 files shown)


idioms/mem-replace.md
```diff
@@ -0,0 +1,83 @@
...
+
+`mem::replace` lets us swap out the value, replacing it with something else. In
+this case, we put in an empty `String`, which does not need to allocate. As a
+result, we get the original `name` *as an owned value*. We can wrap this in
+an `Option` or another enum that we can destructure in the next step to put the
+contained values into our mutably borrowed enum.
+
+## Advantages
+
+Look ma, no allocation! Also you may feel like Indiana Jones while doing it.
```
cbreeden: I still remember the day I learned `mem::replace`, and this is exactly how I felt. 😎

idioms/mem-replace.md
```diff
@@ -0,0 +1,83 @@
...
+        MyEnum::A(ref mut name, x) if x == 0 => {
+            // this takes out our `name` and put in an empty String instead
+            // note that empty strings don't allocate
+            Some(mem::replace(name, "".to_string()))
+        }
+        // nothing to do in all other cases
+        _ => None
+    };
+    // the mutable borrow ends here, so we can change `e`
+    if let Some(name) = have_name { *e = MyEnum::B { name } }
```
cbreeden: `MyEnum::B { name: name }` at least until the [field init shorthand RFC](https://github.com/rust-lang/rfcs/pull/1682) stabilizes if accepted.

cbreeden:
Indeed, this can be quite useful.  I like the example a lot.  Do you mind adding a link to the [table of contents](https://github.com/rust-unofficial/patterns/blob/master/README.md)?  I can add a note to the [clone to satisfy the borrow checker](https://github.com/rust-unofficial/patterns/pull/23) PR, or a new GH issue to remind us to update the relevant links here when that anti-pattern is finished.

idioms/mem-replace.md
```diff
@@ -0,0 +1,83 @@
...
+```rust
+use std::mem;
+
+fn a_to_b(e: &mut MyEnum) {
+    // we mutably borrow `e` here. This precludes us from changing it directly
+    // as in `*e = ...`, because the borrow checker won't allow it. Of course
+    // we could just take a reference to `name` and clone that, but why pay an
+    // extra allocation for something we already have?
+    let have_name = match *e {
+        MyEnum::A(ref mut name, x) if x == 0 => {
```
llogiq: Thanks, fixed.

idioms/mem-replace.md
llogiq: True that. I appear to live in the future from time to time.

llogiq:
@cbreeden that'd be great.

PR #33, opened 2016-10-09
Mention Option::take() as an alternative to mem::replace()

partim:
I’m not sure if this mention should go in the Motivation or Discussion sections. Since `Option` is being mentioned in the Motivation, however, I added it right there.

### DIFF
--- a/idioms/mem-replace.md
+++ b/idioms/mem-replace.md
@@ -53,6 +53,11 @@ result, we get the original `name` *as an owned value*. We can wrap this in
 an `Option` or another enum that we can destructure in the next step to put the
 contained values into our mutably borrowed enum.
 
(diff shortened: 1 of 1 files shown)


idioms/mem-replace.md
```diff
@@ -53,6 +53,11 @@ result, we get the original `name` *as an owned value*. We can wrap this in
 an `Option` or another enum that we can destructure in the next step to put the
 contained values into our mutably borrowed enum.
 
+Note, however, that we you are using an `Option` and are replacing its
```
lfairy: Nit: "we you" should be "when you"

partim: Oops. Fixed and hopefully improved a little overall.

lfairy:
:+1:

PR #34, opened 2016-10-09
Fix a few nits.

CasualX:
Change the match guard to be part of the pattern.

Change `"".to_string()` to a more idiomatic `String::new()` to create a new `String` without memory allocation.

Via [reddit](https://www.reddit.com/r/rust/comments/56l2pk/an_underused_idiom_memreplace_to_avoid_clones/d8kbooz?context=1).


nrc:
Thank you!

Issue #4, opened 2015-10-22
Introduction is empty


MattWis:
Could you just copy the first three paragraphs from your recent blog post?

http://www.ncameron.org/blog/design-patterns-in-rust/

intermo:
I used the first two paragraphs. https://github.com/rust-unofficial/patterns/pull/43#issuecomment-256823836

PR #43, opened 2016-10-28
WIP: Add introduction

intermo:
I started a branch to update the introduction. I added the first three paragraphs of the blog post mentioned in issue #4 and I added a bit more as well. Before I continue, I'd like some feedback as far as wording and topics go so I can make it better.

### DIFF
--- a/intro.md
+++ b/intro.md
@@ -2,7 +2,19 @@
 
 ## Design patterns
 
-What are design patterns? What are idioms? Anti-patterns.
+[Design patterns](https://en.wikipedia.org/wiki/Software_design_pattern) are "general reusable solutions to a commonly occurring problem within a given context in software design". Design patterns are a great way to describe some of the culture and 'tribal knowledge' of programming in a language. Design patterns are very language-specific - what is a pattern in one language may be unnecessary in another due to a language feature, or impossible to express due to a missing feature.
+
+If overused, design patterns can add unnecessary complexity to programs (e.g., see `AbstractSingletonProxyFactoryBean`). However, I think they are a great way to share intermediate and advanced level knowledge about a programming language.
+
+So for these reasons, I have been thinking about design patterns in Rust, and in particular those that are unique to Rust or more common in Rust than other languages. I have started a [pattern catalogue for Rust](https://github.com/rust-unofficial/patterns/blob/master/README.md). It covers idioms (small, simpler design patterns), design patterns, and anti-patterns (patterns which you should strive to avoid). There's not a lot there at the moment. I hope to improve it over the next few months. I would love some help with that - if you like writing documentation and fancy describing some design patterns, please send a PR! There is a list of design patterns which need a description in the contents. If you know of other design patterns in Rust and want to add them, or want to improve some of the existing descriptions, that would be awesome.
+
+## Anti-patterns
+
+[Anti-patterns](https://en.wikipedia.org/wiki/Anti-pattern) are solutions to a "recurring problem that is usually ineffective and risks being highly counterproductive". Just as valuable as the knowledge of how to do a thing is how not to do it. Anti-patterns give us great counter-examples to consider relative to design patterns. Anti-patterns are not confined to code. A process can be an anti-pattern.
+
+## Idioms
+
+Idioms are commonly used styles and patterns largely agreed upon by a community. They are guidelines. Idioms do not _have_ to be used, but if you are not writing idiomatic code you should have a good reason for it. Writing idiomatic code allows other developers to understand what is happening because they are familiar with the form that it has.
 
 ## Design patterns in Rust
 
--- a/intro.md
+++ b/intro.md
@@ -2,7 +2,7 @@
 
 When developing programs, we have to solve many problems. A program can be viewed as a solution to a problem. It can also be viewed as a collection of solutions to many different problems. All of these solutions work together to solve a bigger problem.
 
-There are many problems that share the same form. While the details are different, since they have the same form they can be solved using the same methods. [Design patterns](#design-patterns) are methods to solve common problems when writing software. [Anti-patterns](#anti-patterns) are methods to solve these same common problems. However, while design patterns give us benefits, anti-patterns create more problems. [Idioms](#idioms) are guidelines to follow when coding. They are social norms of the community.
+There are many problems that share the same form. While the details are different, since they have the same form they can be solved using the same methods. [Design patterns](#design-patterns) are methods to solve common problems when writing software. [Anti-patterns](#anti-patterns) are methods to solve these same common problems. However, while design patterns give us benefits, anti-patterns create more problems. [Idioms](#idioms) are guidelines to follow when coding. They are social norms of the community. You can break them, but if you do you should have a good reason for it.
 
 ## Design patterns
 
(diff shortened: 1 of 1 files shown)


cbreeden:
Good stuff, I like it so far

intermo:
I removed the third paragraph from the blog post. It sounds like it should be in a README instead of here.

intro.md
```diff
@@ -1,9 +1,161 @@
...
+
 ## Design patterns
 
-What are design patterns? What are idioms? Anti-patterns.
+[Design patterns](https://en.wikipedia.org/wiki/Software_design_pattern) are "general reusable solutions to a commonly occurring problem within a given context in software design". Design patterns are a great way to describe some of the culture and 'tribal knowledge' of programming in a language. Design patterns are very language-specific - what is a pattern in one language may be unnecessary in another due to a language feature, or impossible to express due to a missing feature.
+
+If overused, design patterns can add unnecessary complexity to programs (e.g., see `AbstractSingletonProxyFactoryBean`). However, I think they are a great way to share intermediate and advanced level knowledge about a programming language.
```
FraGag: Should remove "I think".

patterns/about.md
```diff
@@ -0,0 +1,137 @@
...
+- Targets the wrong problem
+- Lacks formal foundations
+- Leads to inefficient solutions
+- Does not differ significantly from other abstractions
+
+# Design patterns in Rust
+
+Rust has many very unique features. These features give us great benefit by removing whole classes of problems.
+
+Rust is an imperative language, but it follows many functional programming paradigms. One of the biggest hurdles to understanding functional programs when coming from an imperative background is the shift in thinking. Imperative programs describe __how__ to do something, whereas declarative programs describe __what__ to do. Let's sum the numbers from 1 to 10 to show this.
```
killercup: So, this paragraph basically begins a sub-chapter on functional vs. imperative programming, which is confusing to me as a sub-chapter to "Design patterns in Rust". Is this meant as an example? Can you add a headline that clarifies this?

patterns/about.md
```diff
@@ -0,0 +1,137 @@
...
+| 10 | 5 | 15 |
+| 15 | 6 | 21 |
+| 21 | 7 | 28 |
+| 28 | 8 | 36 |
+| 36 | 9 | 45 |
+| 45 | 10 | 55 |
+
+## Types
+
+Rust has strong static types. This can be very different than what you are used to if you are coming from a loosely-typed language. Don't worry, though. Once you get used to them, you'll find the types actually make your life easier. Why? Because you are making implicit assumptions explicit.
```
killercup: > Why? Because you are making implicit assumptions explicit.

Nicely said! Can you add a short example?

patterns/about.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Ownership
+
+Ownership is arguably the central concept of Rust. The ownership system is __the__ reason for the safety guarantees in Rust.
```
killercup: I agree it is important to mention, but ultimately, people will be asking themselves: "How does ownership relate to design patterns?" Is there a pattern that already describes this? Maybe you can link to that (or add a TODO).

patterns/about.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Declarative
+
+```rust
+println!("{}", (1..11).fold(0, |a, b| a + b));
+```
+
+Whoa! This is really different! What's going on here? Remember that with declarative programs we are describing __what__ to do, rather than __how__ to do it. `fold` is a function that [composes](https://en.wikipedia.org/wiki/Function_composition) functions. The name is a convention from Haskell.
+
+Here, we are composing functions of addition (this closure: `|a, b| a + b)`) with a range from 1 to 10. The `0` is the starting point, so `a` is `0` at first. `b` is the first element of the range, `1`. `0 + 1 = 1` is the result. So now we `fold` again, with `a = 1`, `b = 2` and so `1 + 2 = 3` is the next result. This process continues until we get to the last element in the range, `10`.
```
killercup: Good explanation. Maybe you can add that this can also be abstracted away into a new method (`sum`).

killercup: Please note that "composing functions" is not really as easy in Rust as in many other functional languages.

refactoring/about.md
```diff
@@ -0,0 +1,9 @@
+# Refactoring
+
+Refactoring is very important in relation to these topics. Just as important as the other topics covered here, is how to take undesirable code and turn it into great code.
```
killercup: > to these topics

Maybe mention them again? I.e., make explicit that refactoring is basically taking "bad" code and applying patterns and idioms (and domain knowledge and common sense) to make it better.

anti_patterns/about.md
```diff
@@ -0,0 +1,3 @@
+# Anti-patterns
+
+[Anti-patterns](https://en.wikipedia.org/wiki/Anti-pattern) are solutions to a "recurring problem that is usually ineffective and risks being highly counterproductive". Just as valuable as the knowledge of how to do a thing is how not to do it. Anti-patterns give us great counter-examples to consider relative to design patterns. Anti-patterns are not confined to code. A process can be an anti-pattern.
```
intermo: Yea, I like that, but I took _the_ out.

> Just as valuable as knowing how to solve a problem, is knowing how not to solve it.

patterns/about.md
intermo: Thank you! I intend to. I didn't label my TODOs yet. I'm going to get that done before my next push.

patterns/about.md
intermo: I also want to have a high-level overview of the concept. Something like the book metaphor. I think it's an important concept to list since design patterns are seen as intermediate and at that level I think it's important to have a strong understanding of it. As far as relating it, I was planning on showing how Rust outright avoids many problems that other languages have to use patterns to get around.

patterns/about.md
intermo: Yea, good point. I was planning on doing something like currying to show how it's more verbose. I was also planning on highlighting the collection methods as that is the more idiomatic way.

patterns/about.md
```diff
@@ -0,0 +1,137 @@
+# Design patterns
+
+[Design patterns](https://en.wikipedia.org/wiki/Software_design_pattern) are "general reusable solutions to a commonly occurring problem within a given context in software design". Design patterns are a great way to describe some of the culture and 'tribal knowledge' of programming in a language. Design patterns are very language-specific - what is a pattern in one language may be unnecessary in another due to a language feature, or impossible to express due to a missing feature.
+
+If overused, design patterns can add unnecessary complexity to programs (e.g., see `AbstractSingletonProxyFactoryBean`). However, they are a great way to share intermediate and advanced level knowledge about a programming language.
+
+## Creational design patterns
```
intermo: These were committed by mistake. I removed it.

patterns/about.md
intermo: That's a good point. I'm going to break it out as another section. Rust has enough functional influence that a section on how to use it in a functional way will be beneficial.

intermo:
Thanks for all your great feedback!

cbreeden:
Do you have ideas for what you would like to see the in `functional` and `refactoring` folder?

intermo:
I was thinking something similar to what exists in the `patterns` folder. So far my ideas are:

- `functional`
    - currying
    - higher-order functions
    - collections
- `refactoring`
    - refactoring from `unwrap` to `Result`
    - refactoring from `enum` to `Option`

pickfire:
The user became ghost, how should this be handled?

simonsan:
@pickfire We will continue discussing about the content of this PR in #117 

Closing this PR.

PR #45, opened 2016-12-18
Add another example to the mem::replace idiom

### LINKED ISSUE #38: Make "mem::replace to keep owned values in changed enums" example more general
The example seems unrealistic because it uses `if let` instead of `match`. The example should use `match` since it is more general. Most of the cases where I would use this pattern would require the use of `match`.

llogiq:
This fixes #38

### DIFF
--- a/idioms/mem-replace.md
+++ b/idioms/mem-replace.md
@@ -36,6 +36,28 @@ fn a_to_b(e: &mut MyEnum) {
 }
 ```
 
+This also works with more variants:
+
+```Rust
+use std::mem;
+
+enum MultiVariateEnum {
+    A { name: String },
+    B { name: String },
+    C,
+    D
+}
+
+fn swizzle(e: &mut MultiVariateEnum) {
+    use self::MultiVariateEnum::*;
+    *e = match *e {
+        A { ref mut name } => B { name: mem::replace(name, String::new()) },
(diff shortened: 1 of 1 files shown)


idioms/mem-replace.md
```diff
@@ -36,6 +36,28 @@ fn a_to_b(e: &mut MyEnum) {
...
+use std::mem;
+
+enum MultiVariateEnum {
+    A { name: String },
+    B { name: String },
+    C,
+    D
+}
+
+fn swizzle(e: &mut MultiVariateEnum) {
```
nrc: Could you add a comment explaining what this function does and what mem::replace is being used for? This seems somewhat confusing since the code would have the same observable behaviour without the mem::replace.

llogiq: Removing the `mem::replace(..)` requires taking `e` by value, otherwise borrowck will not allow you to take out `name` because it's just borrowed, not owned. One could of course take `e` by value, but there may be reasons against that (e.g. API ergonomics).

llogiq: Will do.

llogiq: Done

lfairy:
Looks good to me. I'll go ahead and merge this.

PR #47, opened 2016-12-21
Deny warnings

### LINKED ISSUE #46: anti pattern: #![deny(warnings)]
Through my work with [clippy](https://github.com/Manishearth/rust-clippy), I've probably had more exposure to the negative consequences than others, so I'm going to write up something in the next few days.

llogiq:
This fixes #46

### DIFF
--- a/anti_patterns/deny-warnings.md
+++ b/anti_patterns/deny-warnings.md
@@ -3,7 +3,7 @@
 ## Description
 
 A well-intentioned crate author wants to ensure their code builds without
-warnings. So they annotate their crate root with
+warnings. So they annotate their crate root with the following:
 
 ## Example
 
@@ -22,28 +22,33 @@ It is short and will stop the build if anything is amiss.
 By disallowing the compiler to build with warnings, a crate author opts out of
 Rust's famed stability. Sometimes new features or old misfeatures need a change
 in how things are done, thus lints are written that `warn` for a certain grace
-period before being turned to `deny`. Sometimes APIs get deprecated, so their
-use will emit a warning where before there was none.
+period before being turned to `deny`.
 
-ALl this conspires to potentially break the build whenever something changes –
-not only for the original developer, but also for those who wish to use their
-code.
+For example, it was discovered that a type could have two `impl`s with the same
+method. This was deemed a bad idea, but in order to make the transition smooth,
+the `overlapping-inherent-impls` lint was introduced to give a warning to those
+stumbling on this fact, before it becomes a hard error in a future release.
+
+Also sometimes APIs get deprecated, so their use will emit a warning where
+before there was none.
+
+All this conspires to potentially break the build whenever something changes.
 
 Furthermore, crates that supply additional lints (e.g. [rust-clippy]) can no
 longer be used unless the annotation is removed.
 
 ## Alternatives
 
-luckily, there are two ways of tackling this problem: First, we can decouple
-the build setting from the code, and second, we can name the lints we want to
-deny explicitly.
+There are two ways of tackling this problem: First, we can decouple the build
+setting from the code, and second, we can name the lints we want to deny
+explicitly.
 
 The following command line will build with all warnings set to `deny`:
  
 ```RUSTFLAGS="-D warnings" cargo build"```
 
 This can be done by any individual developer (or be set in a CI tool like
-travis, but remember that this may break the build when something changes)
+Travis, but remember that this may break the build when something changes)
 without requiring a change to the code.
 
 Alternatively, we can specify the lints that we want to `deny` in the code.
@@ -79,7 +84,8 @@ Here is a list of warning lints that is (hopefully) safe to deny:
 In addition, the following `allow`ed lints may be a good idea to `deny`:
 
 ```rust
-#[deny(missing-docs,
+#[deny(missing-debug-implementations,
+       missing-docs,
        trivial-casts,
        trivial-numeric-casts,
        unused-extern-crates,
@@ -88,7 +94,10 @@ In addition, the following `allow`ed lints may be a good idea to `deny`:
        unused-results)]
 ```
 
-Note that we explicitly did not set the [deprecate attribute]
+Some may also want to add `missing-copy-implementations` to their list.
+
+Note that we explicitly did not add the `deprecated` lint, as it is fairly
+certain that there will be more deprecated APIs in the future.
 
 ## See also
 
--- a/anti_patterns/deny-warnings.md
+++ b/anti_patterns/deny-warnings.md
@@ -88,14 +88,14 @@ In addition, the following `allow`ed lints may be a good idea to `deny`:
        unused-results)]
 ```
 
-Note that we explicitly did not set `deprecated`.
+Note that we explicitly did not set the [deprecate attribute]
 
 ## See also
 
-- [`#[deprecated]`] documentation
+- [deprecate attribute] documentation
 - Type `rustc -W help` for a list of lints on your system. Also type
 `rustc --help` for a general list of options
 - [rust-clippy] is a collection of lints for better Rust code
 
 [rust-clippy]: https://github.com/Manishearth/rust-clippy
-[`#[deprecated]`]: https://doc.rust-lang.org/reference.html#miscellaneous-attributes
+[deprecate attribute]: https://doc.rust-lang.org/reference.html#miscellaneous-attributes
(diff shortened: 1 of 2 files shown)


llogiq:
Oops, somehow messed up the history.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,35 @@
...
+```rust
+#![deny(warnings)]
+
+// All is well.
+```
+
+## Drawbacks
+
+Until, say, the language changes, and as part of the transition, a lint is
+introduced that `warn`s, and Boom! The code no longer builds. Or someone tries
```
nrc: I would phrase this in terms of the Rust language's back compat guarantees - we guarantee (mostly) that we won't break your code with new releases, however, that guarantee does not cover warnings, and thus by using `deny(warnings)` you are opting out of backwards compatibility for your crate.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,35 @@
...
+to compile the code with additional lints (e.g. [clippy]), only to find out
+that they have to remove the annotation first.
+
+## Alternatives
+
+Our plucky developer hero should have used
```
nrc: I would avoid this language, it sounds kinda patronising to me

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,35 @@
...
+`RUSTFLAGS="-D warnings" cargo build"` to build their crate, removing
+the need for the annotation in their code.
+
+## See also
+
+- Type `rustc -W help` for a list of lints on your system. Also type
+`rustc --help` for a general list of options
+- [clippy] is a collection of lints for better Rust code
```
nrc: I wonder if it is worth being explicit and calling it 'rust-clippy' rather than just clippy, since the latter is technically something else (and probably TM etc. MS).

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,35 @@
...
+introduced that `warn`s, and Boom! The code no longer builds. Or someone tries
+to compile the code with additional lints (e.g. [clippy]), only to find out
+that they have to remove the annotation first.
+
+## Alternatives
+
+Our plucky developer hero should have used
+`RUSTFLAGS="-D warnings" cargo build"` to build their crate, removing
+the need for the annotation in their code.
+
```
nrc: I would add some discussion around the fact that usually in Rust, we recommend using in-crate attributes rather than command line flags (several reasons for this, starting with ensuring consistent build experiences without needing a build script on top of Cargo or whatever). However, the advice here is an exception to that rule and maybe explain why it makes sense (which I guess comes down to, deny(warnings) whould be a personal part of the build experience, not something you force onto your clients).

It is probably worth stating the alternative of denying specific warnings as attributes which does not have the behaviour you describe.

lfairy: > It is probably worth stating the alternative of denying specific warnings as attributes which does not have the behaviour you describe.

Is this true in general? Even when only specific warnings are chosen, the meanings of the warnings themselves can change over time. That was part of the motivation for rust-lang/rfcs#1193, after all.

nrc: I would guess it would be 99% true. Warnings very rarely change their meaning. Though they might expand slightly, it is usually to cover edge cases and is unlikely to cause breakage. I expect that any problems this causes is more than outweighed by the benefits of catching errors earlier in the development process and without needing a custom script to run the builds. ISTM that the hazard of the lint case is another lint silently appearing and this approach does not have that problem.

llogiq:
I now rewrote most of the text, added a few allow lints for good measure and fixed history.

@nrc Thank you for the useful comments. I'm now much happier with this PR.

llogiq:
Note to self: Perhaps I should link to the `#[deprecate]` docs, too.

llogiq:
Ok, now I think this is ready to merge. Thoughts?

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
+# `#![deny(warnings)]`
+
+## Description
+
+A well-intentioned crate author wants to ensure their code builds without
+warnings. So they annotate their crate root with
```
lfairy: Reading the rendered view, it's not obvious that the next example finishes this sentence. I'd rather have this written out in full instead.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## Advantages
+
+It is short and will stop the build if anything is amiss.
+
+## Drawbacks
+
+By disallowing the compiler to build with warnings, a crate author opts out of
+Rust's famed stability. Sometimes new features or old misfeatures need a change
```
lfairy: I would like an example of such a change -- any lint from the `future-incompatible` group would be okay.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+in how things are done, thus lints are written that `warn` for a certain grace
+period before being turned to `deny`. Sometimes APIs get deprecated, so their
+use will emit a warning where before there was none.
+
+ALl this conspires to potentially break the build whenever something changes –
```
lfairy: Nit: "ALl" should be "All"

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+not only for the original developer, but also for those who wish to use their
+code.
```
lfairy: This isn't true: Cargo passes `--cap-lints` to rustc, which disables lints while building dependencies. See rust-lang/rfcs#1193.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+
+Furthermore, crates that supply additional lints (e.g. [rust-clippy]) can no
+longer be used unless the annotation is removed.
+
+## Alternatives
+
+luckily, there are two ways of tackling this problem: First, we can decouple
```
lfairy: I think the word "luckily" should be removed. These articles tend to have a sparse writing style, which discourages extra words like this.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+the build setting from the code, and second, we can name the lints we want to
+deny explicitly.
+
+The following command line will build with all warnings set to `deny`:
+ 
+```RUSTFLAGS="-D warnings" cargo build"```
+
+This can be done by any individual developer (or be set in a CI tool like
+travis, but remember that this may break the build when something changes)
```
lfairy: Nit: "Travis" should be capitalized

anti_patterns/deny-warnings.md
llogiq: added.

anti_patterns/deny-warnings.md
```diff
@@ -0,0 +1,101 @@
...
+In addition, the following `allow`ed lints may be a good idea to `deny`:
+
+```rust
+#[deny(missing-docs,
+       trivial-casts,
+       trivial-numeric-casts,
+       unused-extern-crates,
+       unused-import-braces,
+       unused-qualifications,
+       unused-results)]
```
llogiq: done.

llogiq:
Thanks @lfairy for the thorough read-through. This is becoming quite good now.

@nrc @lfairy r?

lfairy:
Looks good to me! I'll wait for any comments from @nrc before merging.

anti_patterns/deny-warnings.md
llogiq: I've changed this to reflect your suggestion.

anti_patterns/deny-warnings.md
llogiq: You're right. I was a bit aggressive because I've had to deal with so many crates doing this, just creating an obstacle for me to test clippy on their code...it's all good now :smile:

anti_patterns/deny-warnings.md
llogiq: All right. I think the list of lints is now complete and usable.

anti_patterns/deny-warnings.md
llogiq: Done.

llogiq:
@nrc ping?

nrc:
LGTM, thanks for the changes!

PR #48, opened 2016-12-31
new idiom: Default trait

llogiq:
Please review. Thank you!

### DIFF
--- a/idioms/default.md
+++ b/idioms/default.md
@@ -1,0 +1,50 @@
+# The `Default` Trait
+
+## Description
+
+Many types in Rust have a [Constructor]. However, this is *specific* to the
+type; Rust cannot abstract over "everything that has a `new()` method". To
+allow this, the [`Default`] trait was conceived, which can be used with
+containers and other generic types (e.g. see [`Option::unwrap_or_default()`]).
+Notably, some containers already implement it where applicable.
+
+Not only do one-element containers like `Cow`, `Box` or `Arc` implement
+`Default` for contained `Default` types, one can automatically
+`#[derive(Default)]` for structs whose fields all implement it, so the more
+types implement `Default`, the more useful it becomes.
+
+## Example
+
+```rust
+// note that we can simply auto-derive Default here.
+#[derive(Default)]
+struct Interesting {
+    /// integers and floats default to zero
+    cardinality: usize,
+    /// bool defaults to false
+    really: bool,
+    /// collections default to empty
+    see_also: Vec<Interesting>
+}
+
+fn interestingness(i: &Interesting) -> usize {
+    if really {
+        i.cardinality
+    } else {
+        // we can use `Option`'s `unwrap_or_default()` method here because our
+        // `Interesting` struct implements `Default`.
+        i.see_also.first().unwrap_or_default().cardinality
+    }
+}
+```
+
+## See also
+
+- The [Constructor] idiom is another way to generate instances that may or may
+not be "default"
+- The [`Default`] documentation (scroll down for the list of implementors)
+- [`Option::unwrap_or_default()`]
+
+[Constructor]: ctor.md
+[`Default`]: https://docs.rust-lang.org/doc/std/default/trait.Default.html
+[`Option::unwrap_or_default()`]: https://docs.rust-lang.org/doc/std/option/enum.Option.html#method.unwrap_or_default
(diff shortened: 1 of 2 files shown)


idioms/default.md
```diff
@@ -0,0 +1,50 @@
...
+    /// integers and floats default to zero
+    cardinality: usize,
+    /// bool defaults to false
+    really: bool,
+    /// collections default to empty
+    see_also: Vec<Interesting>
+}
+
+fn interestingness(i: &Interesting) -> usize {
+    if really {
```
mzji: Should this `really` be `i.really` ?

idioms/default.md
```diff
@@ -0,0 +1,50 @@
+# The `Default` Trait
+
+## Description
+
+Many types in Rust have a [Constructor]. However, this is *specific* to the
```
lfairy: Nit: "constructor" should be lowercase here, at least.

I'm not sure if the Constructor *pattern* should be capitalized or not, but AFAIK when we talk about a constructor *function* it is always in lowercase.

idioms/default.md
```diff
@@ -0,0 +1,50 @@
...
+`Default` for contained `Default` types, one can automatically
+`#[derive(Default)]` for structs whose fields all implement it, so the more
+types implement `Default`, the more useful it becomes.
+
+## Example
+
+```rust
+// note that we can simply auto-derive Default here.
+#[derive(Default)]
+struct Interesting {
```
lfairy: `Default` is often used for configuration structs -- I'd prefer if we use a more configuration-like theme for the example.

idioms/default.md
llogiq: Fixed. Thank you.

idioms/default.md
```diff
@@ -0,0 +1,50 @@
+# The `Default` Trait
+
+## Description
+
+Many types in Rust have a [Constructor]. However, this is *specific* to the
+type; Rust cannot abstract over "everything that has a `new()` method". To
+allow this, the [`Default`] trait was conceived, which can be used with
+containers and other generic types (e.g. see [`Option::unwrap_or_default()`]).
+Notably, some containers already implement it where applicable.
```
nrc: Worth noting, though probably further down, is that `new` and `default` do different things - `new` typically takes values for all or most fields (i.e., is shorthand for the literal, modulo private fields), whereas `default` takes no arguments and thus creates an object 'from nothing'.

idioms/default.md
```diff
@@ -0,0 +1,50 @@
...
+    if i.really {
+        i.cardinality
+    } else {
+        // we can use `Option`'s `unwrap_or_default()` method here because our
+        // `Interesting` struct implements `Default`.
+        i.see_also.first().unwrap_or_default().cardinality
+    }
+}
+```
+
```
nrc: Another detail worth elaborating on is when `Default` can be derived, vs when it must be manually implemented.

idioms/default.md
```diff
@@ -0,0 +1,50 @@
...
+## See also
+
+- The [Constructor] idiom is another way to generate instances that may or may
+not be "default"
+- The [`Default`] documentation (scroll down for the list of implementors)
+- [`Option::unwrap_or_default()`]
+
```
nrc: shameless plug for `derive(new)` - https://crates.io/crates/derive-new/

killercup:
Some talk around new and Default with @llogiq and yours truely: [1](https://github.com/killercup/scribbles/commit/c5863ebdbf343e14d0b5f9e25633ceac6945a41f#commitcomment-19523203) and [2](https://github.com/killercup/scribbles/commit/628f5310cbb601a141dca1dfb057508c9460b006) 😄

sanmai-NL:
Should this PR be merged still? Can we take the last paragraph https://github.com/rust-lang-nursery/api-guidelines/blob/c308779a6b645d1783c1e689cd3286a73ab5a0dc/src/interoperability.md#types-eagerly-implement-common-traits-c-common-traits into consideration (referring to that, not repeating stuff, perhaps expanding that and slimming down this)?

llogiq:
@nrc @lfairy I think I adressed all your requests, so what's keeping this from being merged?

nrc:
I don't think I ever saw the new commits, sorry.

llogiq:
No need to be sorry, I've been swamped with work recently, too.

Issue #44, opened 2016-10-28
Add github pages integration and mdbook for easy publishing and viewing


Michael-F-Bryan:
I've got a [personal project](https://github.com/Michael-F-Bryan/cheat-sheets) which does something pretty similar to this. It's essentially a git repository full of markdown pages. 

What I did is use something called [mdbook](https://github.com/azerupi/mdBook/) which will take all the markdown files and turn them into structured html. I then added a `travis.yml` which will rebuild the html every time I commit some work and then automatically push to the ghpages branch.

I was wondering if you wanted to do something similar to make it easier for people to read and navigate all the useful documents here. It shouldn't require much effort, you just put all the `*.md` source files in a `src` directory and write a short `SUMMARY.md` file to tell mdbook where everything goes. 

here's my `SUMMARY.md`:

```
# Summary

- [Using this Document](./using-the-wiki.md)
- [JavaScript](./JavaScript/javascript.md)
- [Docker](./docker-images.md)
- [Python](./Python/index.md)
    - [SqlAlchemy](./Python/sqlalchemy.md)
- [Basic Server Setup](./basic-server-setup.md)
- [Rust](./Rust/index.md)
    - [Using Rust outside Rust](./Rust/rust_interop.md)
    - [Emails and Templates](./Rust/emails_and_templates.md)
- [Testing In C](./testing_in_c.md)
- [LaTeX Snippets](./latex.md)
```

And the generated documents are viewable [here](https://michael-f-bryan.github.io/cheat-sheets/).

lfairy:
We should consider the original [GitBook](https://www.gitbook.com/). It has built-in GitHub integration (so no fiddling with Travis), has a slick interface, and it's open source as well. Given that we're low on manpower, I'd prefer the solution that takes the least effort to set up and maintain -- GitBook is the best in that regard.

killercup:
Dupe of #6?

lfairy:
Yep, I'd call it a dupe. Thanks for the pointer.

PR #51, opened 2017-01-12
Fixing code examples, statement about working tests on code examples

### LINKED ISSUE #49: rustdoc --test the examples
I have a local branch here that does this. Obviously some examples are not meant to run (e.g. partial copies from `std`), those can be marked with `ignore`.

Otherwise examples should run if it doesn't make them unreadable. I consider adding one or two lines of `use` declarations acceptable.

Thoughts?

llogiq:
This doctests all code examples. There is some fallout to either ignore examples (if I deemed them too far from runnable) or change them to work, or at least compile.

~~I also added a `.travis.yml` so we can use CI to check PRs.~~

fixes #49


nrc:
Hmm, I wonder how worthwhile this is - it found one bug (which is nice), but I feel like every example which gets changed has got more cluttered and less easily understandable and that doesn't feel like a great trade-off to me.

lfairy:
Rustdoc lets you hide setup code by prefixing it with `#`, but I'm not sure how that'll help in our case.

llogiq:
I see two ways going forward:

* write a `build script` that skips lines starting with `#` within code blocks, publish to github-pages
* create an `examples` subdirectory containing complete runnable examples that can be playpen-linked, link them from the inline-examples.

Perhaps we can even do both and have the script build the `examples`?

lfairy:
See also #6 and #44. It would be nice to publish this to a standalone website one day.

Having a separate `examples` directory doesn't feel right, somehow. I think it's because the examples are so short that splitting them out into separate files won't be worth the extra effort. Maintaining a separate set of examples may raise the barrier to entry for contributors as well.

Also, since [rustbook](https://github.com/steveklabnik/rustbook) is built on top of rustdoc, I think it'll strip `#` lines already.

llogiq:
The idea is to have playground links. I'm not sure if this requires a canonical location per example or if the code gets embedded as link data.

lfairy:
Okay, I see what you're aiming for.

The official Rust book has playground links -- e.g. see the section on [lifetimes](https://doc.rust-lang.org/book/lifetimes.html). It looks like they [use the `#` prefix solution](https://github.com/rust-lang/rust/blob/master/src/doc/book/lifetimes.md), displaying the pretty code in the book but linking to the full code on the playground.

Unless there are further issues with their approach, I'm keen to just copy what they do.

llogiq:
Agreed. I'll look into that. Perhaps I should pare down this PR to the one fix I found then.

pickfire:
Ping

simonsan:
fixes #49

simonsan:
Small feedback (imho): I think it would be better at this stage to use Github Actions for that instead of Travis.

lambda-fairy:
mdbook has a built-in doctest feature as well, which would make this PR redundant.

https://rust-lang.github.io/mdBook/cli/test.html

simonsan:
> mdbook has a built-in doctest feature as well, which would make this PR redundant.
> 
> https://rust-lang.github.io/mdBook/cli/test.html

Thanks for the heads-up, indeed, that looks like it would make it redundant when we convert this into a `mdbook`.

simonsan:
@llogiq Could you please remove the `travis.yml` and rebase on master so we can rename the PR and focus on the other fixes you made? Thank you :-)

simonsan (approved):

simonsan:
Seems good to me, @MarcoIeni any additions?

simonsan:
Hey @llogiq is there any reason you're closing this? There are still changes to be merged from your PR that are not the concerning the doctests or Travis, I already edited your branch to remove the parts and changed the title.

simonsan:
Added the statement of `working code examples` to the `template.md` in #123 . What about the additions to the examples and `rust, ignore`? Do we still want to merge that from this PR?

Issue #6, opened 2015-10-23
Convert this to a gitbook/mdbook


dumindu:
Hi,
Thanks for sharing. This will be more readable if you can create a gitbook with this :)

killercup:
I agree. Gitbook's structure also works with https://github.com/steveklabnik/rustbook

mkpankov:
Please note that Gitbook will choke on Rust's doctests or some other parts that require embedded markdown.

That's not a problem if you don't plan to use these features in this project.

vinipsmaker:
> more readable

And ePUB support is always appreciated (comes for free once you use gitbook).

> I agree. Gitbook's structure also works with https://github.com/steveklabnik/rustbook

Does rustbook support ePUB and PDF output like gitbook?

DaseinPhaos:
I'm not pretty sure where this is going but how about converting to mdbook?

PR #53, opened 2017-02-15
Converting to mdbook

DaseinPhaos:
As discussed in [issue#6](https://github.com/rust-unofficial/patterns/issues/6), this PR tries to add rendering and testing support for the project with [mdbook](https://github.com/azerupi/mdBook).

To maintain the current structure of the project while playing nicely with mdbook's building system, a dirty script was added into the project. Currently, just type `cargo run render` in the shell, the rendered pages would be ready for publishing in a subdirectory called `book/`. Check it out on [gh-pages](https://daseinphaos.github.io/patterns/).

Currently two problems remains: 
- testing with `cargo run test` would fail, 
- introductions to subcategories are empty.

Hopefully both problems would be gone after their related PRs being merged into master.

Related: #51 #43


simonsan (commented):
I tested it and I assume this PR might be outdated due to feature additions to `mdbook`. I added a `book.toml` in #111 and it basically does the same without any changes to the repository structure (besides the additions of a `Readme.md` file for each subfolder/subcategory). So imho we could close this and continue work in #111 @MarcoIeni , your thoughts?

MarcoIeni:
I close this in favor of #111 since I think that a lot of the work done for this PR is not required anymore with newer versions of mdbook.
By the way thanks DaseinPhaos for your work!

PR #61, opened 2017-11-09
Use Self in impl

andre-richter:
Saves you time and searching when refactoring the name of the original Struct.

### DIFF
--- a/patterns/builder.md
+++ b/patterns/builder.md
@@ -18,11 +18,11 @@ struct FooBuilder {
 }
 
 impl FooBuilder {
-    fn new(...) -> FooBuilder {
+    fn new(...) -> Self {
         // Set the minimally required fields of Foo.
     }
 
-    fn named(mut self, name: &str) -> FooBuilder {
+    fn named(mut self, name: &str) -> Self {
         // Set the name on the builder itself, and return the builder by value.
     }
 


nrc:
I'm not sure that this is more idiomatic - although you're right that it makes refactoring easier, that should be automated anyway. For reading the code I think this makes things harder because when looking at a function in isolation, if the return type is `Self` you have to find the `impl` declaration to know what type is being returned (i.e., this adds an indirection for the reader).

andre-richter:
I see your point. It certainly makes a difference how many lines of code the `impl` keyword is away from the function in question.
Out of curiosity, was the `Self` keyword introduced to Rust for any other purposes than `Don't Repeat Yourself` in the case mentioned?

softprops:
I switched to Self return types a while back specifically for the types of refactoring it enables. I also feel like its an improvement in ergonimics and adds clarity in semantics. In particular I've found builder style impls gain some clarity when returning Self. The ergonimics in refactorability kick in for types that have type parametization that tends to get. Any type parameter would need repeating n times for method that return the type. If you later decide this type param is not needed. That's n places you have to go back and remove. I've found that Self just plain allows me to move faster as a result.

nrc:
> Out of curiosity, was the Self keyword introduced to Rust for any other purposes than Don't Repeat Yourself in the case mentioned?

In traits, `Self` has a distinct meaning (the type of the implementer). All other uses of `Self` are there really to just allow uniformity.

softprops:
> was the Self keyword introduced to Rust for any other purposes than Don't Repeat Yourself in the case mentioned?

I feel that it also has some parity with the `self`/`&self`/`&mut self` receiver argument of all non associated impl functions.

pickfire:
@andre-richter This is used in some parts, it is easier to write but not easier to read, sometimes you may be looking at a (or multiple) huge impl(s) and you don't know what is the output, it would be hard to find out what is `Self`.

andre-richter:
I forgot this is still idle 

Closing since no consensus reached.

simonsan:
[Doug Milford says in his Rust Tutorial series in the `Impl` section](https://youtu.be/gi0AQ78diSA?t=611) that it's a convention these days?

Marking this as `Zombie` to continue discussion at a later stage.

MarcoIeni:
I try to use it as much as possible.

As nrc was saying:
- cons: readability
- pro: refactoring

We could add this idiom by mentioning both of them so that the reader can get its own idea.

simonsan:
> 
> 
> I try to use it as much as possible.
> 
> As nrc was saying:
> 
>     * cons: readability
> 
>     * pro: refactoring
> 
> 
> We could add this idiom by mentioning both of them so that the reader can get its own idea.

That seems like a really good idea, would you like to write that? ;-)

MarcoIeni:
Created https://github.com/rust-unofficial/patterns/issues/145

Anyway I agree that for examples in this book it is better to not use `Self`, because we prefer readability.

PR #68, opened 2018-02-05
Graphs Pattern

jbgriesner:
An idiomatic way to implement graphs in Rust.

### DIFF
--- a/patterns/graphs.md
+++ b/patterns/graphs.md
@@ -1,0 +1,93 @@
+# Graphs
+
+## Description
+
+Implement a graph data structure that is dynamically modifiable during runtime.
+
+
+## Example
+
+```rust
+// A node of the graph.
+pub struct Node<T> {
+    parent: Option<NodeId>,
+    previous_sibling: Option<NodeId>,
+    next_sibling: Option<NodeId>,
+    first_child: Option<NodeId>,
+    last_child: Option<NodeId>,
+
+    // The actual data stored.
+    pub data: T,
+}
+
+// indexes from the vector used for creating the graph.
+pub struct NodeId {
+    index: usize,
+}
+
+// A memory management arena that ensures that every element of the arena has the same lifetime.
+pub struct Arena<T> {
+    nodes: Vec<Node<T>>,
+}
+
+impl<T> Arena<T> {
+    // Method to create a new node.
+    pub fn new_node(&mut self, data: T) -> NodeId {
+        // Get the next free index
+        let next_index = self.nodes.len();
+
+        // Push the node into the arena
+        self.nodes.push(Node {
+            parent: None,
+            first_child: None,
+            last_child: None,
+            previous_sibling: None,
+            next_sibling: None,
+            data: data,
+        });
+
+        // Return the node identifier
+        NodeId { index: next_index }
+    }
+}
+```
+
+
+## Motivation
+
+This pattern should be used when we need data structures that can be modified during runtime.
+
+
+## Advantages
+
+- Clear and easy to use.
+- Reduce the lifetime complexity within the structures.
+- Avoid runtime borrow checks.
+- Every element within the arena has the same lifetime.
+- Multi processing is possible given that parts of a vector can be shared across threads safely.
+
+
+## Discussion
+
+Region-based memory management is the common method used to deal with graph data structures in Rust.
+This avoids to use interior mutability with data structures as follows:
+
+```
+struct Node<T> {
+    previous: Rc<RefCell<Box<Node<T>>>>,
+    next: Vec<Rc<RefCell<Box<T>>>>,
+    data: T,
+    // ...
+}
+```
+
+which is hard to understand and will also lead into runtime borrow checks.
+
+
+## See also
+
+https://en.wikipedia.org/wiki/Region-based_memory_management
+https://github.com/saschagrunert/indextree
+https://doc.rust-lang.org/1.1.0/rustc_data_structures/graph/
+https://github.com/bluss/petgraph
+https://github.com/danigm/rust-graph-example
--- a/patterns/graphs.md
+++ b/patterns/graphs.md
@@ -20,6 +20,16 @@ pub struct Node<T> {
     pub data: T,
 }
 
+struct Graph<T> {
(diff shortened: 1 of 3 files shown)


patterns/graphs.md
```diff
@@ -0,0 +1,97 @@
...
+
+Implement a graph data structure that is dynamically modifiable during runtime.
+
+
+## Example
+
+```rust
+// A node of the graph.
+pub struct Node<T> {
+    parent: Option<NodeId>,
```
FraGag: If a node can have at most one parent, then the data structure is a *tree*, not a graph. Nodes in a graph can have more than one parent.

nrc:
This is a way to have a graph in Rust, I'm not sure if its the best way. It's also possible to use `Rc` and `RefCell` with some kind of cycle tracking/detection, or unsafe code in numerous ways.

patterns/graphs.md
lfairy: It looks more like a *forest*, since this code allows for multiple roots 😉

jbgriesner:
I made a distinction.
I will code different ways to have graph.

lambda-fairy:
Found this while going through old PRs.

In my view, the intention of this patterns repo is to document situations where there is usually one "best" solution. So an overview of many different graph implementations, while valuable, would not belong here.

I will close this PR. Thanks!

simonsan:
> In my view, the intention of this patterns repo is to document situations where there is usually one "best" solution. So an overview of many different graph implementations, while valuable, would not belong here.
> 
> I will close this PR. Thanks!

Imho a `best` solution is developed over time, I think it could be valuable, as this content doesn't exist in the repository already, to state explicitly that it's not sure to be the best solution but that PRs are welcome to improve the examples or encourage discussion on it to come to a better solution, so I'm reopening it for further inspection.

patterns/graphs.md
```diff
@@ -0,0 +1,116 @@
...
+
+https://en.wikipedia.org/wiki/Region-based_memory_management
+
+https://github.com/saschagrunert/indextree
+
+https://doc.rust-lang.org/1.1.0/rustc_data_structures/graph/
+
+https://github.com/bluss/petgraph
+
+https://github.com/danigm/rust-graph-example
```
simonsan, line 116:
```suggestion
- [Region-based memory management](https://en.wikipedia.org/wiki/Region-based_memory_management)
- [Module rustc_data_structures::graph](https://doc.rust-lang.org/1.1.0/rustc_data_structures/graph/)
- [Code example: Arena based tree structure with multithreading support](https://github.com/saschagrunert/indextree)
- [Code example: Graph data structure library](https://github.com/bluss/petgraph)
- [Code example: How to implement complex data structure in rust](https://github.com/danigm/rust-graph-example)
```

patterns/graphs.md
```diff
@@ -0,0 +1,115 @@
...
+which is hard to understand and will also lead into runtime borrow checks.
+
+
+## See also
+
+- [Region-based memory management](https://en.wikipedia.org/wiki/Region-based_memory_management)
+- [Module rustc_data_structures::graph](https://doc.rust-lang.org/1.1.0/rustc_data_structures/graph/)
+- [Code example: Arena based tree structure with multithreading support](https://github.com/saschagrunert/indextree)
+- [Code example: Graph data structure library](https://github.com/bluss/petgraph)
+- [Code example: How to implement complex data structure in rust](https://github.com/danigm/rust-graph-example)
```
simonsan, line 115:
```suggestion
- [Code example: How to implement complex data structure in Rust](https://github.com/danigm/rust-graph-example)
```

simonsan:
But yeah, I know what lambda-fairy wanted to state, it's hard to tell this is a good solution or this not. My guess with the book would probably be a subchapter with graphs and then give some examples 

or

probably better to explain the internal graph data structures as a pattern, something like that:
https://doc.rust-lang.org/1.1.0/src/rustc_data_structures/graph/mod.rs.html#11-403 

and give input on how you would do that on not give some custom implementation. But maybe some recommended crates with already implemented data structures to use, to not reinvent the wheel.

Hmm, thoughts?

fade2black:
I think Graph data structure, on its own, is too wide topic to cover in a couple of pages. As a rule, complexity of algorithms on graphs directly depends their [representation](https://en.wikipedia.org/wiki/Graph_(abstract_data_type)#Representations). What kind of representation we will choose depends on our use case. Factors to take into account: 
- type of graphs: complete graph, trees, directed, undirected, etc.
- compute intensive: speed/complexity of algorithms is important
- scarce memory (embedded systems for example)
- what kind of algorithm we implement: parallel, distributed...
- Is graph dynamic?: we constantly add/remove vertices/edges...
- ... 

Let me give a simple example. Say, I know my graph changes (new edges or vertex) once a day, but the graph is almost complete and in my algorithms I traverse neighbours of a single vertex not too often. I'd probably use adjacency matrix.

simonsan:
I think as well, that it should be a complete chapter of different graph structures, would you agree?

fade2black:
@simonsan I agree. We could categorise into use cases or representations, or anything else what people think. But not everything in a single page. Of course the Rust way should be put on the first place.

simonsan:
> 
> 
> @simonsan I agree. We could categorise into use cases or representations, or anything else what people think. But not everything in a single page. Of course the Rust way should be put on the first place.

I opened an issue for discussions, it's easier going forward I guess

pickfire:
This is too wide and there are few patterns to solve this. But there are also many cases of graph, I think we may need a category rather than writing it like this. For example:

- self-referencing graphs (there are even multiple ways to solve this IIRC and this is a common pitfall I think)
- struct of array vs array of structure
- allocate memory somewhere else and keep reference/index

simonsan:
> 
> 
> This is too wide and there are few patterns to solve this. But there are also many cases of graph, I think we may need a category rather than writing it like this. For example:
> 
>     * self-referencing graphs (there are even multiple ways to solve this IIRC and this is a common pitfall I think)
> 
>     * struct of array vs array of structure
> 
>     * allocate memory somewhere else and keep reference/index

Let's use issue #202 to discuss it, I'll mark this as zombie for the future and close it for now.

PR #69, opened 2018-02-08
Fix wording in "Fold" pattern

king6cong:


patterns/fold.md
```diff
@@ -100,7 +100,7 @@ exclusively, the original copy of the data structure cannot be re-used. On the
 other hand if a node is not changed, reusing it is very efficient.
 
 If we were to operate on borrowed references, the original data structure can be
-reused, however, if a node is unchanged it must be cloned, which can be
+reused, however, a node must be cloned even not changed, which can be
```
lambda-fairy, line 103: There needs to be a conjunction between the "even" and "not".

Suggest:

```suggestion
reused; however, a node must be cloned even if unchanged, which can be
```

lambda-fairy (dismissed):

simonsan (approved):

simonsan:
Thank you!

PR #75, opened 2018-11-21
new pattern: Contain Unsafety in Small Modules

### LINKED ISSUE #74: Pattern: contain unsafety in small module
When dealing with `unsafe` code, the containing module has to ensure safety by upholding invariants. To make this feasible, the module should be as small as possible, only containing the unsafe functionality and the code necessary to uphold the guarantees, and embedding this in a larger module that can safely use the abstraction.

This keeps the code needed for an unsafe audit in manageable size.

llogiq:
This closes #74 .

### DIFF
--- a/patterns/unsafe-mods.md
+++ b/patterns/unsafe-mods.md
@@ -1,4 +1,4 @@
-# Contain Unsafety in Small Modules
+# Contain unsafety in small modules
 
 ## Description
 
@@ -16,7 +16,7 @@ If you have `unsafe` code, create the smallest possible module that can uphold t
 
 ## Examples
 
-* The [`optional`](https://crates.io/crates/optional) crate has one unsafe module for converting a `&T` to either a slice of one `[T]` or an empty slice.
+* The [`toolshed`](https://docs.rs/toolshed) crate contains its unsafe operations in submodules, presenting a safe interface to users.
 * `std`s `String` class is a wrapper over `Vec<u8>` with the added invariant that the contents must be valid UTF-8. The operations on `String` ensure this behavior. However, users have the option of using an `unsafe` method to create a `String`, in which case the onus is on them to guarantee the validity of the contents.
 
 ## See also
--- a/patterns/unsafe-mods.md
+++ b/patterns/unsafe-mods.md
@@ -17,7 +17,7 @@ If you have `unsafe` code, create the smallest possible module that can uphold t
 ## Examples
 
 * The [`optional`](https://crates.io/crates/optional) crate has one unsafe module for converting a `&T` to either a slice of one `[T]` or an empty slice.
-* `std`s `String` class is basically just a wrapper over `Vec<u8>` with the added invariant that the contents must be valid unicode. The operations on `String` ensure this behavior, however, users have the option of using an `unsafe` method to create a `String`, in which case the onus is on them to guarantee the validity of the contents.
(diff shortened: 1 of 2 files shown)


patterns/unsafe-mods.md
```diff
@@ -0,0 +1,24 @@
...
+* Writing the outer module is much easier, since you can count on the guarantees of the inner module
+
+## Disadvantages
+
+* Sometimes, it may be hard to find a suitable interface.
+* The abstraction may introduce inefficiencies.
+
+## Examples
+
+* The [`optional`](https://crates.io/crates/optional) crate has one unsafe module for converting a `&T` to either a slice of one `[T]´ or an empty slice.
```
FraGag: Wrong tick:

```suggestion
* The [`optional`](https://crates.io/crates/optional) crate has one unsafe module for converting a `&T` to either a slice of one `[T]` or an empty slice.
```

llogiq: Thank you!

patterns/unsafe-mods.md
```diff
@@ -0,0 +1,24 @@
...
+* Writing the outer module is much easier, since you can count on the guarantees of the inner module
+
+## Disadvantages
+
+* Sometimes, it may be hard to find a suitable interface.
+* The abstraction may introduce inefficiencies.
+
+## Examples
+
+* The [`optional`](https://crates.io/crates/optional) crate has one unsafe module for converting a `&T` to either a slice of one `[T]` or an empty slice.
```
lfairy: There's already a [standard library function](https://doc.rust-lang.org/std/slice/fn.from_ref.html) for doing that, so I suggest using another example :stuck_out_tongue:

patterns/unsafe-mods.md
```diff
@@ -0,0 +1,24 @@
...
+* `std`s `String` class is basically just a wrapper over `Vec<u8>` with the added invariant that the contents must be valid unicode. The operations on `String` ensure this behavior, however, users have the option of using an `unsafe` method to create a `String`, in which case the onus is on them to guarantee the validity of the contents.
```
lfairy: * I think the words "basically just" don't add value and can be removed
* It's more precise to say "UTF-8" than "Unicode"
* I think the comma before "however" should be a period or semicolon instead, as per [this guide](http://www.chompchomp.com/terms/conjunctiveadverb.htm).

```suggestion
* `std`s `String` class is a wrapper over `Vec<u8>` with the added invariant that the contents must be valid UTF-8. The operations on `String` ensure this behavior. However, users have the option of using an `unsafe` method to create a `String`, in which case the onus is on them to guarantee the validity of the contents.
```

patterns/unsafe-mods.md
llogiq: Yeah, I wrote that before the function became available. The newest version has no more `unsafe`.

I'll look around for other examples.

patterns/unsafe-mods.md
llogiq: Thanks, you are completely right. This reads much better now.

patterns/unsafe-mods.md
llogiq: Found one – toolshed keeps all unsafety in submodules.

llogiq:
@lfairy what is needed to merge this?

lfairy:
@llogiq just a ping from you 😉

Sorry about the delay. I'll merge this.

PR #79, opened 2019-03-05
Update vec.rs path

NealEhardt:
vec.rs has moved https://github.com/rust-lang/rust/blob/master/src/liballoc/vec.rs

Relevant commit from 2017 https://github.com/rust-lang/rust/commit/eadda7665eb31b1e7cb94a503b4d5cf5c75474c0#diff-541a0d1eb7d496a1f91ed729f52261ed


nrc:
Thanks!

PR #80, opened 2019-04-09
RAII: stands for "Resource Acquisition is Initialisation"

xakon:
Unfortunately, the name is even more terrible than initially was thought!


nrc:
Thanks!

PR #90, opened 2019-11-23
Make builder pattern more practical

TGotwig:
https://github.com/TGotwig/patterns/blob/patch-1/patterns/builder.md


nesteruk:
It is confusing to have `bar: String` in both `Foo` and `FooBuilder` — consider renaming one of the `bar`s to something else.

TGotwig:
@nesteruk Changed this and a little bit more 🙂

patterns/builder.md
```diff
@@ -8,38 +8,53 @@ Construct an object with calls to a builder helper.
...
+    bar: String,
 }
 
-struct FooBuilder {
+// Optional, define the textual representation for Foo.
+impl fmt::Display for Foo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "bar: {}", self.bar)
+    }
+}
```
lambda-fairy, line 23: This `Display` impl feels irrelevant to illustrating the builder pattern. Can we replace this with `#[derive(Debug)]`, and use `format!("{:?}", foo)` in the test instead?

lambda-fairy (dismissed):
LGTM, modulo the comment below

patterns/builder.md
TGotwig: I will look for this soon 🙂

pickfire:
This example does not seemed to be explaining much, maybe we could have a real-life example or an explanation on why it needs to be constructed this way rather than just specifying the struct directly?

patterns/builder.md
TGotwig: Yeah you are right! Changed this up 👷

TGotwig:
@pickfire Hmm, changed it a little bit up.
I think that runnable code is better than not runnable code and the comments should explain enough 🧐

patterns/builder.md
```diff
@@ -2,46 +2,51 @@
...
+
+  pub fn name(mut self, bar: &str) -> Builder {
+    // Set the name on the builder itself, and return the builder by value.
+    self.builder_bar = bar.to_string();
+    self
+  }
+
+  // If we can get away with not consuming the Builder here, that is an
+  // advantage. It means we can use the builder as a template for constructing many Foos.
+  pub fn finish(self) -> Foo {
```
pickfire, line 37: Isn't it usually `build`?

pickfire (commented):
The indentation seemed off, should be 4 spaces rather than 2 spaces.

patterns/builder.md
```diff
@@ -2,46 +2,51 @@
...
+  // Lots of complicated fields.
+  bar: String,
 }
 
-struct FooBuilder {
-    // Probably lots of optional fields.
-    ...
+pub struct Builder {
+  // Probably lots of optional fields.
+  builder_bar: String,
```
pickfire, line 18: We can probably just keep the same `bar` here.

patterns/builder.md
```diff
@@ -2,46 +2,49 @@
...
 ```rust
-struct Foo {
+#[derive(Debug)]
+pub struct Foo {
     // Lots of complicated fields.
+    bar: String,
 }
 
-struct FooBuilder {
+pub struct Builder {
```
pickfire, line 16: Why shouldn't we keep the builder name as `FooBuilder`?

patterns/builder.md
```diff
@@ -2,46 +2,49 @@
...
-    // various aspects of a Foo.
-    ...
-
     // If we can get away with not consuming the Builder here, that is an
-    // advantage. It means we can use the builder as a template for constructing
-    // many Foos.
-    fn finish(&self) -> Foo {
-        // Create a Foo from the FooBuilder, applying all settings in FooBuilder to Foo.
+    // advantage. It means we can use the builder as a template for constructing many Foos.
+    pub fn finish(self) -> Foo {
```
pickfire, line 37: Some parts seemed inconsistent, some uses `finish` and below uses `build`.

TGotwig: Unfortunately I had this overwritten, but now it is `build` again 👷

patterns/builder.md
TGotwig: On second thought you are right, changed it back ↩️

patterns/builder.md
```diff
@@ -2,46 +2,51 @@
...
     }
+  }
 }
 
-fn main() {
-    let f = FooBuilder::new().named("Bar").with_attribute(...).finish();
+#[test]
+fn builder_test() {
+  let foo: Foo = Builder::new().name("Y").finish();
+  assert_eq!(format!("{:?}", foo), "Foo { bar: \"Y\" }");
```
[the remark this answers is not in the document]
TGotwig: Changed this up 👷

patterns/builder.md
```diff
@@ -2,46 +2,58 @@
...
+pub struct Foo {
     // Lots of complicated fields.
+    bar: String,
+}
+
+impl PartialEq for Foo {
+    fn eq(&self, other: &Self) -> bool {
+        self.bar == other.bar
+    }
 }
```
pickfire, line 20: Why not just do derive `PartialEq` above? Rather than writing this from scratch?

patterns/builder.md
```diff
@@ -2,46 +2,58 @@
...
 
-fn main() {
-    let f = FooBuilder::new().named("Bar").with_attribute(...).finish();
+#[test]
+fn builder_test() {
+    let foo = Foo {
+        bar: String::from("Y"),
+    };
+    let foo_from_builder: Foo = FooBuilder::new().name("Y").build();
```
pickfire, line 54: Do we need this type signature, isn't it inferred?

patterns/builder.md
```diff
@@ -2,46 +2,58 @@
...
         // Set the minimally required fields of Foo.
+        FooBuilder {
+            bar: "X".to_string(),
+        }
     }
 
-    fn named(mut self, name: &str) -> FooBuilder {
+    pub fn name(mut self, bar: &str) -> FooBuilder {
         // Set the name on the builder itself, and return the builder by value.
+        self.bar = bar.to_string();
```
pickfire, line 37: Is this a good pattern to clone within the function when users can provide `String`? I am not quite sure how this is done in the wild.

patterns/builder.md
TGotwig: It's not needed but best practice to do so, I think 🤔

pickfire: I am not sure if it is best practice but for docs examples (including rust), most types are explicitly state to let others to understand easily but I think this example should be way to simple, if they get to this page they should already be able to see through this simple example.

pickfire (approved):
Looks good to me. I think it would be good to have one of the maintainer to check this though.

simonsan:
@TGotwig Hey, thank you for the PR. Could you please rebase on `master` so the CI is checking this through while we get to reviewing this PR? Thank you ;-)

TGotwig:
@simonsan Sure, will do! 😃

simonsan:
> @simonsan Sure, will do! 😃

Sorry, haven't seen that there is no addition to summary or anything needed. So I could do it myself. ;-) It's fine then.

simonsan:
@MarcoIeni You also want to look over again?

MarcoIeni (approved):

simonsan:
Thanks! :-)

TGotwig:
Cool! 😊🚀

PR #93, opened 2020-03-05
Show call of default()

kellerkindt:
The code lacks an example call with ```default()```. It might not be completely clear to a Rust beginner, what the ```#[derive(Default)]``` actually enabled.


lambda-fairy:
Thanks!

PR #94, opened 2020-03-24
New idiom: Out of the Box Dynamic Dispatch

llogiq:


idioms/ootb-dyn-dispatch.md
```diff
@@ -0,0 +1,83 @@
...
+## Description
+
+We can dynamically dispatch over multiple values, however, to do so, we need
+to declare multiple variables to bind differently-typed objects. To extend the
+lifetime as necessary, we can use deferred conditional initialization, as seen
+below:
+
+## Example
+
+``rust
```
0x20F: Codeblocks start with three ```

twe4ked:
Great write up! I was a bit confused by the title at first, perhaps something along the lines of “Dynamic Dispatch without Box<T>”?

llogiq:
Perhaps on-stack dynamic dispatch?

twe4ked:
That sounds good to me ☺️

README.md
```diff
@@ -24,6 +24,7 @@ language.
 * [Pass variables to closure](idioms/pass-var-to-closure.md)
 * [`mem::replace(_)` to avoid needless clones](idioms/mem-replace.md)
 * [Temporary mutability](idioms/temporary-mutability.md)
+* [Out of the Box Dynamic Dispatch](idioms/ootb-dyn-dispatch.md)
```
Nashenas88: Isn't this title now out of sync with the title below?

llogiq: Right, fixed.

llogiq:
It's been a while since there was any movement on this. Is there something I should do or is this good to merge?

llogiq:
@lambda-fairy @Manishearth anyone?

lambda-fairy:
Hey @llogiq!

Sorry, I've had way too many things going on lately 😢 

It looks like the document is still called `ootb-dyn-dispatch.md`. Can you please rename that to something like `on-stack-dyn-dispatch.md`?

Otherwise it looks good to me 👍

burdges:
I suppose LLVM always eliminates the duplicate stack usage here, so there is never a reason to do this with a union.

pickfire:
@llogiq I think you should also add an example for `Option`.

idioms/ootb-dyn-dispatch.md
```diff
@@ -0,0 +1,83 @@
+# On-Stack Dynamic Dispatch
```
pickfire, line 1: I think this is more well-known as out-of-box dynamic dispatch from blog posts, maybe we could mention it as well?

llogiq: I don't think that will be necessary.

llogiq:
@pickfire I've noted in the "see also" section that `Option.as_ref()` solves most of the same problem for us.

llogiq:
Moved the file and rebased. @lambda-fairy r?

PR #95, opened 2020-04-25
TODO's

Ben-PH:
I propose that these two additional TODOs will be of value once implemented, and while I'm doing something, clarifying that a topic is templated might be useful


lambda-fairy:
Thanks!

Issue #11, opened 2015-10-23
Address issues from the reddit thread


nrc:
https://www.reddit.com/r/rust/comments/3ptlgq/design_patterns_in_rust_blog_post/

simonsan:
Big topic, we need to work through that thread and check against already existing issues/PRs and the status quo of the content.

simonsan:
I read through it and haven't really found anything that would be worth opening with, also it's really old (5 years) and probably makes sense to cut it at this point. Closing this now. ;-)

PR #102, opened 2020-10-31
Adding Nick Camerons talk at PDXRust 2016 to additional resources section

simonsan:
Thought might be helpful for people just looking at this repository but not knowing about the talk.


simonsan:
Probably better in a `Materials` section to read more about some stuff rather than randomly stated in the doc. We could add the reddit thread from the #11 and some blog posts as well, plus the API guidelines.

@MarcoIeni What do you think?

simonsan:
With the book already working we could probably add a small `materials` section or `further reading` (or something like this) in another md-file and include it in the TOC

MarcoIeni:
> With the book already working we could probably add a small `materials` section or `further reading` (or something like this) in another md-file and include it in the TOC

sounds great!

simonsan:
> 
> 
> > With the book already working we could probably add a small `materials` section or `further reading` (or something like this) in another md-file and include it in the TOC
> 
> sounds great!
> 
> Maybe `Additional materials`?
> 
> I don't like `further reading` because it will be a section with also videos, not only written stuff.

Japp, `Additional materials` sounds better to me as well

additional_materials.md
```diff
@@ -0,0 +1,7 @@
+# Additional materials
+
+A collection of complementary helpful content
+
+## Talks
+
+- ["Design Patterns in Rust" by Nick Cameron at the PDRust (2016)](https://www.youtube.com/watch?v=Pm_oO0N5B9k)
```
MarcoIeni, line 7:
```suggestion
- [Design Patterns in Rust](https://www.youtube.com/watch?v=Pm_oO0N5B9k) by Nick Cameron at the PDRust (2016)
```

MarcoIeni: Do you think it's better like this? Anyway, feel free to keep this or your version and then you can merge it!

simonsan: I think I like that your version better. But you need to approve again after a change ;-)

simonsan:
> 
> 
> > With the book already working we could probably add a small `materials` section or `further reading` (or something like this) in another md-file and include it in the TOC
> 
> sounds great!
> 
> Maybe `Additional materials`?

Hmm, thinking of it, what do you think of `Additional Resources`?

MarcoIeni (dismissed):

MarcoIeni (approved):

PR #106, opened 2020-12-06
Several FFI Sections: Strings, Errors, and API Design

jhwgh1968:
This is still in draft because it's early, but I wanted to open this PR to claim it and show where I'm going. Let me know if this is the right approach, or if I missed conversation occurring elsewhere and jumped the gun.

Looking through the PRs and the Issues, it looked like no one was tackling FFI. Well, I volunteer!

It's a big subject, of course, so I'm only offering one small part: exporting Rust type APIs to other languages, using object-based API design.

The example may seem rather obscure, but I think it's a good illustration. It is also one I am very familiar with. I learned Rust by trying to port the QDBM library from C (before finding something off-the-shelf). I got about half way done, and learned a _lot_ about C bindings and memory safety! :smile: I figure this is the best way to capture a lot of that knowledge, and share it with the world.


jhwgh1968:
That turned into more typing than I was expecting! But I think it's all pretty good -- or at least, worth reviewing. If it's better to break this up a bit, I can do that as well.

simonsan:
@jhwgh1968 Hey there, nice work! As a first step could you rebase your branch on `master` please and link the corresponding new files in `SUMMARY.md`? We'll review this PR the upcoming days.

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,80 @@
...
+    };
+    crate::log(msg_str, level);
+}
+```
+
+## Advantages
+
+The example is is written to ensure that:
+1. The `unsafe` block is as small as possible.
+1. The pointer with an "untracked" lifetime becomes a "tracked" shared reference
```
MarcoIeni, line 40:
```suggestion
2. The pointer with an "untracked" lifetime becomes a "tracked" shared reference
```

There are a lot of lists that are like this (1. 1. 1.)

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,80 @@
...
+
+When accepting strings via FFI through pointers, there are two principles that should be followed:
+1. Keep foreign strings "borrowed", rather than copying them directly.
+2. Minimize `unsafe` code during the conversion.
+
+## Motivation
+
+Rust has built-in support for C-style strings with its `CString` and `CStr` types. However, there are different approaches one can take with strings that are being accepted from a foreign caller of a Rust function.
+
+The best practice is simple: use `CStr` in such a way as to minimize unsafe code, and created a borrowed slice. If an owned String is needed, call `to_string()` on the string slice.
```
MarcoIeni, line 13:
```suggestion
The best practice is simple: use `CStr` in such a way as to minimize unsafe code, and create a borrowed slice. If an owned String is needed, call `to_string()` on the string slice.
```

MarcoIeni:
CI is failing, can you fix that? Thanks!

simonsan:
@jhwgh1968  Not sure how well FFI and doc tests work, if there is an example that you can't get through doc test please mark it with 

```
```rust,ignore
```

idioms/ffi-accepting-strings.md
jhwgh1968: This was intentional.

Both `mdbook` and markdown renumber lists in the same block regardless of the actual numbers you put in. Given this, I thought it would be easier to write `1. 1. 1.` than `1. 2. 3.` to avoid future changes that rendered okay ending up `1. 2. 2. 3.` in the source, which might cause some scratching of heads.

That said, I will still fix this if you would like.

jhwgh1968:
Thanks for the tip, @simonsan. Although I did end up putting `ignore` on most blocks, the CI did catch a couple typos and silly things first!

Fingers crossed, CI should pass on the version I just pushed. Once it does, then, I'll get to the other minor comments.

idioms/ffi-intro.md
```diff
@@ -0,0 +1,5 @@
+# FFI Idioms
+
+Writing FFI code is an entire course in itself. However, there are several idioms here that can act a pointers, and avoid traps for inexperienced users of unsafe Rust.
+
+This section contains idioms that may be useful when doing FFI.
```
simonsan, line 5: Proposing to in-repo link here the sections as a small TOC if it works as an intro to those files

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
...
+1. Make the lifetime of owned strings as long as possible.
+1. Minimize `unsafe` code during the conversion.
+1. If the C code can modify the string data, use `Vec` instead of `CString`.
+1. Unless the FFAPI requires it, the ownership of the string should not transfer to the callee.
+
+## Motivation
+
+Rust has built-in support for C-style strings with its `CString` and `CStr` types. However, there are different approaches one can take with strings that are being sent to a foreign function call from a Rust function.
+
+The best practice is simple: use `CString` in such a way as to minimize unsafe code. However, a secondary caveat is that *the object must live long enough*, meaning the lifetime should be maximized. In addition, the documentation explains that "round-tripping" a `CString` after modification is UB, so additional work is necessary in that case.
```
simonsan, line 16: Will be easier for changes to follow due to VCS caveats. Also I propose to use `unsafe` instead of unsafe in the complete PR to make sure that it stands out to the reader.

```suggestion
The best practice is simple: use `CString` in such a way as to minimize `unsafe` code. 
However, a secondary caveat is that *the object must live long enough*, meaning the lifetime should be maximized.
In addition, the documentation explains that "round-tripping" a `CString` after modification is UB, so additional work is necessary in that case.
```

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
...
+fn get_error_from_ffi() -> Result<String, std::ffi::IntoStringError> {
+    let mut buffer = vec![0u8; 1024];
+    unsafe {
+        // SAFETY: calling an FFI whose documentation implies that the input need only live as long
+        // as the call
+        let written: usize = geterr(buffer.as_mut_ptr(), 1023).into();
+        buffer.truncate(written + 1);
+    }
+    std::ffi::CString::new(buffer).unwrap().into_string()
+}
```
simonsan, line 46: Same here, let's wrap that in a module

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+## Advantages
+
+The example is is written to ensure that:
```
simonsan, line 51:
```suggestion
The example is written in a way to ensure that:
```

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
...
+A common mistake (so common it's in the documentation) is to not use the variable in the first block:
+
+```rust,ignore
+fn report_error<S: Into<String>>(err: S) -> Result<(), std::ffi::NulError> {
+    unsafe {
+        // SAFETY: whoops, this contains a dangling pointer!
+        seterr(std::ffi::CString::new(err.into())?.as_ptr());
+    }
+    Ok(())
+}
```
simonsan, line 65: Like that maybe?

```suggestion
// mod unsafe_module {
  // other module content
  
  fn report_error<S: Into<String>>(err: S) -> Result<(), std::ffi::NulError> {
      unsafe {
          // SAFETY: whoops, this contains a dangling pointer!
          seterr(std::ffi::CString::new(err.into())?.as_ptr());
      }
      Ok(())
  }
}
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+1. All Encapsulated types should be *owned* by Rust, *managed* by the user, and *opaque*.
+1. All Transactional data types should be *owned* by the user, and *transparent*.
+1. All library behavior should be functions acting upon Encapsulated types.
+1. All library behavior should be encapsulated into types not based on structure, but *provenance/lifetime*.
+
+## Motivation
+
+Rust has built-in FFI support to other languages. It does this by providing a way for crate authors to provide C-compatible APIs through different ABIs (though that is unimportant to this practice).
+
+Well-designed Rust FFI follows C API design principles, while compromising the design in Rust as little ass possible. There are three goals with any foreign API:
```
simonsan, line 16:
```suggestion
Well-designed Rust FFI follows C API design principles, while compromising the design in Rust as little as possible. There are three goals with any foreign API:
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+1. Make it easy to use in the target language.
+1. Avoid the API dictating internal unsafety on the Rust side as much as possible.
+1. Keep the potential for memory unsafety and Rust UB as small as possible.
```
simonsan, line 20:
```suggestion
1. Keep the potential for memory unsafety and Rust `undefined behaviour` as small as possible.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+Rust code must trust the memory safety of the foreign language beyond a certain point. However, every bit of `unsafe` code on the Rust side is an opportunity for bugs, or to exacerbate UB.
```
simonsan, line 22:
```suggestion
Rust code must trust the memory safety of the foreign language beyond a certain point. However, every bit of `unsafe` code on the Rust side is an opportunity for bugs, or to exacerbate `undefined behaviour`.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+For example, if a pointer provenance is wrong, that may be a segfault due to invalid memory access. But if it is manipulated by unsafe code, it could become full-blown heap corruption.
+
+The Object-Based API design allows for writing shims that have good memory safety characteristics, and a clean boundary of what is safe and what is not.
```
simonsan, line 26:
```suggestion
The Object-Based API design allows for writing shims that have good memory safety characteristics, and a clean boundary of what is safe and what is `unsafe`.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Code Example
+
+The POSIX standard defines the API to access an on-file database, known as DBM [1]. It is an excellent example of an "object-based" API.
```
simonsan, line 30:
```suggestion
The POSIX standard defines the API to access an on-file database, known as [DBM](https://www.mankier.com/0p/ndbm.h). It is an excellent example of an "object-based" API.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+As noted earlier, this type is *transparent* to the user. But also, this type is *owned* by the user. This has subtle ramifications, due to that pointer inside it. The question is, who owns the memory that pointer points to?
+
+The answer for best memory safety is, "the user". But in cases such as retrieving a value, the user does not know how to allocate it correctly (since they don't know how long the value is). In this case, the library code is expected to use the heap that the user has access to -- such as the C library `malloc` and `free` -- and then *transfer ownership* in the Rust sense.
+
+This may all seem speculative, but this is what a pointer means in C. It means the same thing as Rust: "user defined lifetime." The user of the library needs to read the documentation in order to use it correctly. That said, there are some decisions that have fewer or greater consequences if users do it wrong. Minimizing those is what this best practice is about, and the key is to *transfer ownership of everything that is transparent*.
+
+[1] https://www.mankier.com/0p/ndbm.h
```
simonsan, line 69:
```suggestion
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+For example, if a pointer provenance is wrong, that may be a segfault due to invalid memory access. But if it is manipulated by unsafe code, it could become full-blown heap corruption.
+
+The Object-Based API design allows for writing shims that have good memory safety characteristics, and a clean boundary of what is safe and what is not.
+
+## Code Example
+
+The POSIX standard defines the API to access an on-file database, known as DBM [1]. It is an excellent example of an "object-based" API.
+
+Here is the definition in C, which hopefully should be easy to read for those involved in FFI. Commentary below should help explain it for those who miss the subtleties.
```
simonsan, line 32:
```suggestion
Here is the definition in C, which hopefully should be easy to read for those involved in FFI. While the commentary below should help explain it for those who miss the subtleties.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+This API defines two types: `DBM` and `datum`.
+
+The `DBM` type was called an "encapsulated" type above. It is designed to contain internal state, and acts as an entry point for the library's behavior.
+
+It is completely opaque to the user, who cannot create a `DBM` themselves since they don't know its size or layout. Instead, they must call `dbm_open`, and that only gives them *a pointer to one*.
+
+This means all `DBM`s are "owned" by the library in a Rust sense. The internal state of unknown size is kept in memory controlled by the library, not the user. The user can only manage its life cycle with `open` and `close`, and perform operations on it with the other functions.
+
+The `datum` type was called a "transactional" type above. It is designed to facilitate the exchange of information between the library and it user.
```
simonsan, line 57:
```suggestion
The `datum` type was called a "transactional" type above. It is designed to facilitate the exchange of information between the library and its user.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+The answer for best memory safety is, "the user". But in cases such as retrieving a value, the user does not know how to allocate it correctly (since they don't know how long the value is). In this case, the library code is expected to use the heap that the user has access to -- such as the C library `malloc` and `free` -- and then *transfer ownership* in the Rust sense.
+
+This may all seem speculative, but this is what a pointer means in C. It means the same thing as Rust: "user defined lifetime." The user of the library needs to read the documentation in order to use it correctly. That said, there are some decisions that have fewer or greater consequences if users do it wrong. Minimizing those is what this best practice is about, and the key is to *transfer ownership of everything that is transparent*.
+
+[1] https://www.mankier.com/0p/ndbm.h
+
+## Advantages
+
+This minimizes the number of memory safety guarantees the user must uphold to a relatively small number:
+1. Do not call any function with a pointer not return by `dbm_open` (invalid access or corruption).
```
simonsan, line 74:
```suggestion
1. Do not call any function with a pointer not returned by `dbm_open` (invalid access or corruption).
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+1. Do not call any function on a pointer after close (use after free).
+1. The `dptr` on any `datum` must be NULL, or point to a valid slice of memory at the advertised length.
```
simonsan, line 76:
```suggestion
1. The `dptr` on any `datum` must be `NULL`, or point to a valid slice of memory at the advertised length.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+pub extern "C" fn dbm_iter_next(iter: *mut DbmKeysIter, key_out: *const datum) -> libc::c_int {
+    /* THIS API IS A BAD IDEA! For real applications, use object-based design instead. */
+}
+#[no_mangle]
+pub extern "C" fn dbm_iter_del(*mut DbmKeysIter) {
+    /* THIS API IS A BAD IDEA! For real applications, use object-based design instead. */
+}
+```    
+
+This API loses a key piece of information: the lifetime of the iterator must not exceed the lifetime of the `Dbm` object that owns it. A user of the library could which causes the iterator to outlive the data it is iterating on, resulting in reading uninitialized memory.
```
simonsan, line 117:
```suggestion
This API loses a key piece of information: the lifetime of the iterator must not exceed the lifetime of the `Dbm` object that owns it. A user of the library could use it in a way which causes the iterator to outlive the data it is iterating on, resulting in reading uninitialized memory.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+For the C programmers out there, spot the bug:
```
simonsan, line 119:
```suggestion
This example written in C contains a bug that will be explained afterwards:
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+This bug is a classic. Here's what happens when the iterator returns the end-of-iteration marker:
+
+1. The loop condition sets `l` to zero, and enters the loop because `0 >= 0`.
+1. The length is incremented, in this case by zero.
+1. The if statement is true, so the database is closed. There should be a break statement here.
+1. The loop condition executes again, causing a `next` call on the closed object.
+
+The worst part about this bug? If the Rust implementation was careful, this code will work most of the time! If the memory for the `Dbm` object is not immediately reused, an internal check will almost certainly return an error, resulting in the iterator returning a -1. But occasionally, it will be cause a segmentation fault, or even worse nonsensical memory corruption!
```
simonsan, line 155:
```suggestion
The worst part about this bug? If the Rust implementation was careful, this code will work most of the time!
If the memory for the `Dbm` object is not immediately reused, an internal check will almost certainly return an error, resulting in the iterator returning a `-1` indicating an error.
But occasionally, it will cause a segmentation fault, or even worse nonsensical memory corruption!
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+datum   dbm_nextkey(DBM *);
+```
+
+Thus, all of the lifetimes were bound together, and such unsafety was prevented.
+
+## Disadvantages
+
+However, this design choice also has a number of drawbacks, which should be considered as well.
+
+First, the API itself becomes less expressive. With POSIX DBM, there is only one iterator per object, and every call changes its state. This is much more restrictive than iterators in almost any language, even though it is safe. Perhaps with other related objects, whose lifetimes are less heirarchical, this limitation is more of a cost than the safety.
```
simonsan, line 172:
```suggestion
First, the API itself becomes less expressive. With POSIX DBM, there is only one iterator per object, and every call changes its state. 
This is much more restrictive than iterators in almost any language, even though it is safe. 
Perhaps with other related objects, whose lifetimes are less hierarchical, this limitation is more of a cost than the safety.
```

patterns/ffi-intro.md
```diff
@@ -0,0 +1,5 @@
+# FFI Idioms
+
+Writing FFI code is an entire course in itself. However, there are several idioms here that can act a pointers, and avoid traps for inexperienced users of unsafe Rust.
+
+This section contains design patterns that may be useful when doing FFI.
```
simonsan, line 5: Same here, please create a small TOC ;-)

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
+# Type Consolidation into Wrappers
+
+## Description
+
+This pattern is designed to allow gracefully handling multiple related types, while minimizing the surface area for memory unsafety.
+
+One of the cornerstones of Rust's aliasing rules is lifetimes. This ensures many patterns of access between types can be memory safe, including against data races.
```
simonsan, line 7:
```suggestion
One of the cornerstones of Rust's aliasing rules is lifetimes. This ensures that many patterns of access between types can be memory safe, data race safety included.
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+However, when Rust types are exported to other languages, they are usually transformed into pointers. In Rust, a pointer means "the user manages the lifetime of the pointee." It is their responsibility to avoid memory unsafety.
```
simonsan, line 9: `Pointee` wasn't intended I assume?

```suggestion
However, when Rust types are exported to other languages, they are usually transformed into pointers. In Rust, a pointer means "the user manages the lifetime of the pointer." It is their responsibility to avoid memory unsafety/ensure memory safety.
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+Some level of trust in the user code is thus required, notably around use-after-free which Rust can do nothing about. However, some API designs place higher burdens than others on the code written in the other language.
+
+The lowest risk API is the "consolidated wrapper", where all possible interactions with an object are folded into a "wrapper type", while keeping the Rust API clean.
+
+## Code Example
+
+To understand this, let us look at a classic example of an API to export: iteration through a collection. The API looks like this:
```
simonsan, line 17:
```suggestion
To understand this, let us look at a classic example of an API to export: iteration through a collection. 

The API looks like this:
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. The iterator is wrapped with the collection, and is initialized with `first_key`.
+2. Each call to `next_key` will advance the iterator.
+3. Calls to `next_key` if the iterator is at the end will do nothing.
+
+If the iterator implements `nth()` efficiently, then it is possible to make it ephemeral to each function call:
+
+```rust,ignore
```
simonsan, line 25: Is this needed?

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+            self.iter_next += 1;
+            Some(next)
+        } else {
+            None
+        }
+    }
+}
+```
+
+As a result, the wrapper is simple and contains no unsafe code.
```
simonsan, line 47:
```suggestion
As a result, the wrapper is simple and contains no `unsafe` code.
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+```rust,ignore
+struct MySetWrapper {
+    myset: MySet,
+    iter_next: usize,
+    // created from a transmuted Box<KeysIter + 'self>
+    iterator: Option<NonNull<KeysIter<'static>>>,
+}
+```
+
+With `transmute` being used to extend a lifetime, and a pointer to hide it, it's ugly already. But it gets even worse: *any other operation can cause Rust UB*.
```
simonsan, line 76:
```suggestion
With `transmute` being used to extend a lifetime, and a pointer to hide it, it's ugly already. But it gets even worse: *any other operation can cause Rust `undefined behaviour`*.
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+        &mut (*myset).myset
+    };
+
+    /* ...check and cast key and value data... */
+
+    match myset.store(casted_key, casted_value) {
+        Ok(_) => 0,
+        Err(e) => e.into()
+    }
+}
```
simonsan, line 95: Let's wrap it in a module again ;-)

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+If the iterator exists when this function is called, we have violated one of Rust's aliasing rules. According to Rust, the mutable reference in this block must have *exclusive* access to the object. If the iterator simply exists, it's not exclusive, so we have UB![1](#UB-footnote)
```
simonsan, line 98:
```suggestion
If the iterator exists when this function is called, we have violated one of Rust's aliasing rules. According to Rust, the mutable reference in this block must have *exclusive* access to the object. If the iterator simply exists, it's not exclusive, so we have `undefined behaviour`![1](#UB-footnote)
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+To avoid this, we must have a way of ensuring that mutable reference really is exclusive. That basically means clearing out the iterator's shared reference while it exists, and then reconstructing it. In most cases, that will still be less efficient than the C version.
+
+Some may ask: how can C do this more efficiently? The answer is, it cheats. Rust's aliasing rules are the problem, and C simply ignores them for its pointers. In exchange, it is common to see code that is declared in the manual as "not thread safe" under some or all circumstances. In fact, [The GNU C library has an entire lexicon dedicated to concurrent behavior!](https://manpages.debian.org/buster/manpages/attributes.7.en.html)
```
simonsan, line 102:
```suggestion
Some may ask: how can C do this more efficiently? The answer is, it cheats. Rust's aliasing rules are the problem, and C simply ignores them for its pointers.
In exchange, it is common to see code that is declared in the manual as "not thread safe" under some or all circumstances.
In fact, [The GNU C library has an entire lexicon dedicated to concurrent behavior!](https://manpages.debian.org/buster/manpages/attributes.7.en.html)
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+Rust would rather make everything memory safe all the time, for both safety and optimizations that C code cannot attain. Being denied access to certain shortcuts is the price Rust programmers pay.
```
simonsan, line 104:
```suggestion
Rust would rather make everything memory safe all the time, for both safety and optimizations that C code cannot attain. Being denied access to certain shortcuts is the price Rust programmers need to pay.
```

simonsan (changes requested):
Nice read! Though here are some suggestions and changes ;-)

patterns/ffi-intro.md
```diff
@@ -0,0 +1,5 @@
+# FFI Idioms
```
simonsan, line 1:
```suggestion
# FFI Patterns
```

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
+# Passing Strings
+
+## Description
+
+When passing strings to FFI functions, there are four principles that should be followed:
+
+1. Make the lifetime of owned strings as long as possible.
+1. Minimize `unsafe` code during the conversion.
+1. If the C code can modify the string data, use `Vec` instead of `CString`.
+1. Unless the FFAPI requires it, the ownership of the string should not transfer to the callee.
```
simonsan, line 10:
```suggestion
1. Unless the Foreign Function API requires it, the ownership of the string should not transfer to the callee.
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Advantages
+
+This makes APIs safer to use, avoiding issues with lifetimes between types. See [Object-Based APIs](./ffi-exports.md) for more on the advantages and pitfalls this avoids.
+
+## Disadvantages
+
+Often, wrapping types is quite difficult, and sometimes a Rust API compromise would make things easier.
+
+As an example, consider an iterator which does not efficiently implement `nth()`. It would definitely be worth putting in special logic to make the object handle iteration internally, or to support a different access pattern efficiently that only the FFAPI will use.
```
simonsan, line 57:
```suggestion
As an example, consider an iterator which does not efficiently implement `nth()`. It would definitely be worth putting in special logic to make the object handle iteration internally, or to support a different access pattern efficiently that only the Foreign Function API will use.
```

idioms/ffi-accepting-strings.md
MarcoIeni: I have never seen this, nice to know.
Maybe there are some people who reads from markdown directly (for example if you want to contribute).
I will propose the changes, so you only have to accept it :)

idioms/ffi-errors.md
```diff
@@ -0,0 +1,108 @@
+# Error Handling in FFI
+
+## Description
+
+In foreign languages like C, errors are represented by return codes. However, Rust's type system allows much more rich error information to be captured a propogated through a full type.
+
+This best practice shows different kinds of error codes, and how to expose them in a usable way:
+1. Flat Enums should be converted to integers and returned as codes.
+1. Structured Enums should be converted to an integer code with a string error message for detail.
+1. Custom Error Types should become "transparent", with a C representation.
```
MarcoIeni, line 10:
```suggestion
2. Structured Enums should be converted to an integer code with a string error message for detail.
3. Custom Error Types should become "transparent", with a C representation.
```

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,74 @@
...
+    std::ffi::CString::new(buffer).unwrap().into_string()
+}
+```
+
+## Advantages
+
+The example is is written to ensure that:
+1. The `unsafe` block is as small as possible.
+1. The `CString` lives long enough.
+1. Errors with typecasts are always propagated when possible.
```
MarcoIeni, line 54:
```suggestion
2. The `CString` lives long enough.
3. Errors with typecasts are always propagated when possible.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
+# Object-Based APIs
+
+## Description
+
+When designing APIs in Rust which are exposed to other languages, there are some important design principles which are contrary to normal Rust API design:
+
+1. All Encapsulated types should be *owned* by Rust, *managed* by the user, and *opaque*.
+1. All Transactional data types should be *owned* by the user, and *transparent*.
+1. All library behavior should be functions acting upon Encapsulated types.
+1. All library behavior should be encapsulated into types not based on structure, but *provenance/lifetime*.
```
MarcoIeni, line 10:
```suggestion
2. All Transactional data types should be *owned* by the user, and *transparent*.
3. All library behavior should be functions acting upon Encapsulated types.
4. All library behavior should be encapsulated into types not based on structure, but *provenance/lifetime*.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+```
+
+This bug is a classic. Here's what happens when the iterator returns the end-of-iteration marker:
+
+1. The loop condition sets `l` to zero, and enters the loop because `0 >= 0`.
+1. The length is incremented, in this case by zero.
+1. The if statement is true, so the database is closed. There should be a break statement here.
+1. The loop condition executes again, causing a `next` call on the closed object.
```
MarcoIeni, line 153:
```suggestion
2. The length is incremented, in this case by zero.
3. The if statement is true, so the database is closed. There should be a break statement here.
4. The loop condition executes again, causing a `next` call on the closed object.
```

simonsan:
@jhwgh1968 Please don't force-push so we can actually keep commit history, we'll squash it at the end ;-)

patterns/ffi-intro.md
```diff
@@ -0,0 +1,5 @@
...
+
+Writing FFI code is an entire course in itself. However, there are several idioms here that can act a pointers, and avoid traps for inexperienced users of unsafe Rust.
```
simonsan, line 3:
```suggestion
Writing FFI code is an entire course in itself. However, there are several idioms here that can act as pointers, and avoid traps for inexperienced users of unsafe Rust.
```

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,80 @@
...
+        }
+    };
+    crate::log(&msg_str, level);
+}
+```
+
+This code in inferior to the original in two respects:
+
+1. There is much more `unsafe` code, and more importantly, more invariants it must uphold.
+2. Due to the extensive arithmetic required, there is a bug in this version that cases Rust UB.
```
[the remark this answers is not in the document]
simonsan: merged the UB suggestion, but could you still write a small explanation for the bug ;-)

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,76 @@
+# Passing Strings
+
+## Description
+
+When passing strings to FFI functions, there are four principles that should be followed:
+
+1. Make the lifetime of owned strings as long as possible.
+1. Minimize `unsafe` code during the conversion.
+1. If the C code can modify the string data, use `Vec` instead of `CString`.
+1. Unless the Foreign Function API requires it, the ownership of the string should not transfer to the callee.
```
simonsan, line 10:
```suggestion
1. Make the lifetime of owned strings as long as possible.
2. Minimize `unsafe` code during the conversion.
3. If the C code can modify the string data, use `Vec` instead of `CString`.
4. Unless the Foreign Function API requires it, the ownership of the string should not transfer to the callee.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Motivation
+
+Rust has built-in FFI support to other languages. It does this by providing a way for crate authors to provide C-compatible APIs through different ABIs (though that is unimportant to this practice).
+
+Well-designed Rust FFI follows C API design principles, while compromising the design in Rust as little as possible. There are three goals with any foreign API:
+
+1. Make it easy to use in the target language.
+1. Avoid the API dictating internal unsafety on the Rust side as much as possible.
+1. Keep the potential for memory unsafety and Rust UB as small as possible.
```
simonsan:
```suggestion
1. Make it easy to use in the target language.
2. Avoid the API dictating internal unsafety on the Rust side as much as possible.
3. Keep the potential for memory unsafety and Rust UB as small as possible.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Motivation
+
+Rust has built-in FFI support to other languages. It does this by providing a way for crate authors to provide C-compatible APIs through different ABIs (though that is unimportant to this practice).
+
+Well-designed Rust FFI follows C API design principles, while compromising the design in Rust as little as possible. There are three goals with any foreign API:
+
+1. Make it easy to use in the target language.
+1. Avoid the API dictating internal unsafety on the Rust side as much as possible.
+1. Keep the potential for memory unsafety and Rust `undefined behaviour` as small as possible.
```
simonsan, line 20:
```suggestion
1. Make it easy to use in the target language.
2. Avoid the API dictating internal unsafety on the Rust side as much as possible.
3. Keep the potential for memory unsafety and Rust `undefined behaviour` as small as possible.
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,180 @@
...
+
+This may all seem speculative, but this is what a pointer means in C. It means the same thing as Rust: "user defined lifetime." The user of the library needs to read the documentation in order to use it correctly. That said, there are some decisions that have fewer or greater consequences if users do it wrong. Minimizing those is what this best practice is about, and the key is to *transfer ownership of everything that is transparent*.
+
+[1] https://www.mankier.com/0p/ndbm.h
+
+## Advantages
+
+This minimizes the number of memory safety guarantees the user must uphold to a relatively small number:
+1. Do not call any function with a pointer not return by `dbm_open` (invalid access or corruption).
+1. Do not call any function on a pointer after close (use after free).
```
simonsan, line 74:
```suggestion
2. Do not call any function on a pointer after close (use after free).
```

idioms/ffi-passing-strings.md
```diff
@@ -0,0 +1,76 @@
...
+        // SAFETY: whoops, this contains a dangling pointer!
+        seterr(std::ffi::CString::new(err.into())?.as_ptr());
+    }
+    Ok(())
+}
+```
+
+This code will result in a dangling pointer, because the lifetime of the `CString` is not extended by the pointer creation, unlike if a reference were created.
+
+Another issue frequently raised is that the initialization of a 1k vector of zeroes is "slow". However, recent versions of Rust actually optimize that particular macro to a call to `zmalloc`, meaning it is as fast as the operating system's ability to return zeroed memory (which is quite fast).
```
simonsan, line 72:
```suggestion
Another issue frequently raised is that the initialization of a 1k vector of zeroes is "slow". 
However, recent versions of Rust actually optimize that particular macro to a call to `zmalloc`, meaning it is as fast as the operating system's ability to return zeroed memory (which is quite fast).
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,183 @@
...
+However, this design choice also has a number of drawbacks, which should be considered as well.
+
+First, the API itself becomes less expressive. With POSIX DBM, there is only one iterator per object, and every call changes its state. 
+This is much more restrictive than iterators in almost any language, even though it is safe. 
+Perhaps with other related objects, whose lifetimes are less hierarchical, this limitation is more of a cost than the safety.
+
+Second, depending on the relationships of the API's parts, significant design effort may be involved. Many of the easier design points have other patterns associated with them:
+
+* [Wrapper Type Consolidation](./ffi-wrappers.md) groups multiple Rust types together into an opaque "object"
+* [FFI Error Passing](./ffi-errors.md) explains error handling is with integer codes and sentinel return values (such as null pointers)
```
simonsan, line 180:
```suggestion
* [FFI Error Passing](../idioms/ffi-errors.md) explains error handling is with integer codes and sentinel return values (such as `NULL` pointers)
```

patterns/ffi-export.md
```diff
@@ -0,0 +1,183 @@
...
+* [Accepting Foreign Strings](./ffi-accepting-strings.md) allows accepting strings with minimal unsafe code, and is easier to get right than [Passing Strings to FFI](./ffi-passing-strings.md)
```
simonsan, line 181:
```suggestion
* [Accepting Foreign Strings](../idioms/ffi-accepting-strings.md) allows accepting strings with minimal unsafe code, and is easier to get right than [Passing Strings to FFI](../idioms/ffi-passing-strings.md)
```

patterns/ffi-wrappers.md
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+}
+```
+
+As a result, the wrapper is simple and contains no `unsafe` code.
+
+## Advantages
+
+This makes APIs safer to use, avoiding issues with lifetimes between types. See [Object-Based APIs](./ffi-exports.md) for more on the advantages and pitfalls this avoids.
```
simonsan, line 53:
```suggestion
This makes APIs safer to use, avoiding issues with lifetimes between types. See [Object-Based APIs](./ffi-export.md) for more on the advantages and pitfalls this avoids.
```

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,102 @@
...
+
+        crate::log(&msg_str, level);
+    }
+}
+```
+
+This code in inferior to the original in two respects:
+
+1. There is much more `unsafe` code, and more importantly, more invariants it must uphold.
+2. Due to the extensive arithmetic required, there is a bug in this version that cases Rust `undefined behaviour`.
```
simonsan, line 98: @jhwgh1968 Last nitpick from my side, I fixed the other suggestions:

Could you please make a small explanation for the bug as you did here?
https://github.com/rust-unofficial/patterns/pull/106/files#diff-e268dfd7a80ac1b0ef9fb2ed7a903397027a5637b1d9d2c84ebb0476e18803dcR147

simonsan (changes requested):

jhwgh1968:
Overall you definitely improved things for me. Thank you!

I have revised the bug you mentioned, and added the explanation you requested. I also fixed a couple of minor things, related to word choice and grammar, mostly in my original. In case it's helpful, I finally found a way to do footnotes that works, and looks the way I intended.

In my mind, at least, it is ready to merge!

simonsan:
Nice! Will get to it after merging the `markdownlinter`, so we can let it run as well over this as well ;-)

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,106 @@
...
+
+The bug here is a simple mistake in pointer arithmetic: the string was copied, all `msg_len` bytes of it. However, the NUL terminator at the end was not.
+
+The Vector then had its size *set* to the length of the *zero padded string* -- rather than *resized* to it, which could have added a zero at the end. As a result, the last byte in the Vector is uninitialized memory. When the `CString` is created at the bottom of the block, its read of the Vector will cause `undefined behaviour`!
```
simonsan, line 100:
```suggestion
The Vector then had its size *set* to the length of the *zero padded string* -- rather than *resized* to it, which could have added a zero at the end. As a result, the last byte in the Vector is uninitialized memory.
When the `CString` is created at the bottom of the block, its read of the Vector will cause `undefined behaviour`!
```

idioms/ffi-accepting-strings.md
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+```
+
+This code in inferior to the original in two respects:
+
+1. There is much more `unsafe` code, and more importantly, more invariants it must uphold.
+2. Due to the extensive arithmetic required, there is a bug in this version that cases Rust `undefined behaviour`.
+
+The bug here is a simple mistake in pointer arithmetic: the string was copied, all `msg_len` bytes of it. However, the NUL terminator at the end was not.
```
simonsan, line 98:
```suggestion
The bug here is a simple mistake in pointer arithmetic: the string was copied, all `msg_len` bytes of it.
However, the `NUL` terminator at the end was not.
```

idioms/ffi-accepting-strings.md
```diff
@@ -97,6 +95,12 @@ This code in inferior to the original in two respects:
 1. There is much more `unsafe` code, and more importantly, more invariants it must uphold.
 2. Due to the extensive arithmetic required, there is a bug in this version that cases Rust `undefined behaviour`.
 
+The bug here is a simple mistake in pointer arithmetic: the string was copied, all `msg_len` bytes of it. However, the NUL terminator at the end was not.
+
+The Vector then had its size *set* to the length of the *zero padded string* -- rather than *resized* to it, which could have added a zero at the end. As a result, the last byte in the Vector is uninitialized memory. When the `CString` is created at the bottom of the block, its read of the Vector will cause `undefined behaviour`!
+
+Like many such issues, this would be difficult issue to track down. Sometimes it would panic because the string was not UTF-8, sometimes it would put a weird character at the end of the string, sometimes it would just completely crash.
```
simonsan, line 102:
```suggestion
Like many such issues, this would be difficult issue to track down.
Sometimes it would panic because the string was not `UTF-8`, sometimes it would put a weird character at the end of the string, sometimes it would just completely crash.
```

patterns/ffi-export.md
```diff
@@ -29,7 +29,7 @@ The Object-Based API design allows for writing shims that have good memory safet
 
 The POSIX standard defines the API to access an on-file database, known as [DBM](https://web.archive.org/web/20210105035602/https://www.mankier.com/0p/ndbm.h). It is an excellent example of an "object-based" API.
 
-Here is the definition in C, which hopefully should be easy to read for those involved in FFI. While the commentary below should help explaining it for those who miss the subtleties.
+Here is the definition in C, which hopefully should be easy to read for those involved in FFI. The commentary below should help explaining it for those who miss the subtleties.
```
simonsan, line 32:
```suggestion
Here is the definition in C, which hopefully should be easy to read for those involved in FFI.
The commentary below should help explaining it for those who miss the subtleties.
```

patterns/ffi-export.md
```diff
@@ -177,7 +177,7 @@ Perhaps with other related objects, whose lifetimes are less hierarchical, this
 Second, depending on the relationships of the API's parts, significant design effort may be involved. Many of the easier design points have other patterns associated with them:
 
 * [Wrapper Type Consolidation](./ffi-wrappers.md) groups multiple Rust types together into an opaque "object"
```
simonsan, line 179:
```suggestion
- [Wrapper Type Consolidation](./ffi-wrappers.md) groups multiple Rust types together into an opaque "object"
```

patterns/ffi-export.md
```diff
@@ -177,7 +177,7 @@ Perhaps with other related objects, whose lifetimes are less hierarchical, this
...
-* [FFI Error Passing](../idioms/ffi-errors.md) explains error handling is with integer codes and sentinel return values (such as `NULL` pointers)
+* [FFI Error Passing](../idioms/ffi-errors.md) explains error handling with integer codes and sentinel return values (such as `NULL` pointers)
 * [Accepting Foreign Strings](../idioms/ffi-accepting-strings.md) allows accepting strings with minimal unsafe code, and is easier to get right than [Passing Strings to FFI](../idioms/ffi-passing-strings.md)
```
simonsan:
```suggestion
- [FFI Error Passing](../idioms/ffi-errors.md) explains error handling with integer codes and sentinel return values (such as `NULL` pointers)
- [Accepting Foreign Strings](../idioms/ffi-accepting-strings.md) allows accepting strings with minimal unsafe code, and is easier to get right than [Passing Strings to FFI](../idioms/ffi-passing-strings.md)
```

patterns/ffi-wrappers.md
```diff
@@ -107,7 +108,7 @@ pub mod unsafe_module {
 }
 ```
 
-If the iterator exists when this function is called, we have violated one of Rust's aliasing rules. According to Rust, the mutable reference in this block must have *exclusive* access to the object. If the iterator simply exists, it's not exclusive, so we have `undefined behaviour`![1](#UB-footnote)
+If the iterator exists when this function is called, we have violated one of Rust's aliasing rules. According to Rust, the mutable reference in this block must have *exclusive* access to the object. If the iterator simply exists, it's not exclusive, so we have `undefined behaviour`! <sup>[1](#UB-footnote)</sup>
```
simonsan, line 111:
```suggestion
If the iterator exists when this function is called, we have violated one of Rust's aliasing rules.
According to Rust, the mutable reference in this block must have *exclusive* access to the object.
If the iterator simply exists, it's not exclusive, so we have `undefined behaviour`! <sup>[1](#UB-footnote)</sup>
```

simonsan (approved):

simonsan:
Let's ship it! Great job! <3

simonsan:
@jhwgh1968 What's your reddit username? (if you have one)

Michael-F-Bryan (commented):
I hope you don't mind, but I've written down a couple comments and questions that came to me as I was reading.

Unsafe code tends to require a higher level of scrutiny, doubly so for documentation/tutorials around `unsafe`. I've tried to be pedantic about using the correct terminology and patterns, just because we don't want to teach people the wrong thing.

If you disagree on any points, I really want to hear about it. Either I'll be proven wrong and learn something, or we'll improve the overall document, so it's a win-win situation :slightly_smiling_face:

simonsan:
@Michael-F-Bryan Much appreciated! If you both @jhwgh1968 could condense something into a new `pull request` that would be great and really much appreciated. Improvement by iteration, so to say. ;-)

Michael-F-Bryan:
@jhwgh1968 sorry for being late to the party and posting a review after the PR was merged. Would you be interested in making a follow-up?

I've added review comments for most of the points, but there are a couple systemic issues I identified:

- Unwinding across the FFI boundary is UB, so all calls to non-trivial Rust code in FFI functions must be wrapped in `std::panic::catch_unwind()`
- Almost all of your FFI bindings should `unsafe fn`'s because their memory safety depends on the caller doing the right thing, so *not* marking them as `unsafe` is technically unsound
- No null pointer checks
- You can tell some of the code hasn't been compiled and run, for example the `get_error_from_ffi()` function will always panic because you call `CString::new()` with a buffer that contains a null (the terminator)

simonsan:
@Michael-F-Bryan If it's not so comfortable for you both to discuss in this closed and merged PR it might be really best to open a new one with your improvements on the topic I feel? If I could do that I would open this one again to encourage discussing on it, but sadly Github doesn't allow it.

PR #22, opened 2016-05-31
Adding catch_panic anti-pattern

liamzdenek:

PR #109, opened 2021-01-01
Adding catch_panic anti-pattern

simonsan:
Due to inactivity of the author and maintainers not being able to edit the PR I'll move #22 to a new branch in this repository.

Future fixes to the PR will be made on this branch.


pickfire:
Isn't web frameworks doing this? Is this really an anti-pattern?

MarcoIeni:
> Isn't web frameworks doing this? Is this really an anti-pattern?

In the "advantages" section it is mentioned

anti_patterns/catch_panic.md
```diff
@@ -0,0 +1,81 @@
+# catch_unwind for exceptions
+
+## Description
+
+This anti-pattern arises when the method for catching (`panic::catch_unwind`) an
+unexpected problem (implementation bug) is used to handle an expected problem,
+such as a missing file.
```
MarcoIeni, line 7:
```suggestion
This anti-pattern arises when the method for catching ([panic::catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)) an
unexpected problem (implementation bug) is used to handle an expected problem,
such as a missing file.
```

MarcoIeni:
We can take a lot from the official documentation: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html

anti_patterns/catch_panic.md
```diff
@@ -0,0 +1,81 @@
...
+
+## See also
+
+[The Rust Book: Error Handling](https://doc.rust-lang.org/book/error-handling.html)
+
+[Rust 1.9 announcement, which contains a description of this antipattern](http://blog.rust-lang.org/2016/05/26/Rust-1.9.html)
+
+[Result documentation](http://doc.rust-lang.org/std/result/enum.Result.html)
+
+[panic::catch_unwind documentation](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)
```
simonsan, line 81:
```suggestion
- [The Rust Book: Error Handling](https://doc.rust-lang.org/book/error-handling.html)
- [Rust 1.9 announcement, which contains a description of this antipattern](http://blog.rust-lang.org/2016/05/26/Rust-1.9.html)
- [Result documentation](http://doc.rust-lang.org/std/result/enum.Result.html)
- [panic::catch_unwind documentation](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)
```

simonsan:
@MarcoIeni 
Why is it expecting a `+` now for ul? :D

```
anti_patterns/catch_panic.md:76:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
anti_patterns/catch_panic.md:77:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
anti_patterns/catch_panic.md:78:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
anti_patterns/catch_panic.md:79:1 MD004/ul-style Unordered list style [Expected: plus; Actual: dash]
```

MarcoIeni:
By default it expects the style to be consistent, see [this](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004---unordered-list-style)

Since there is a `+` on line 66 it expects that all the lists follow that.

simonsan:
> 
> 
> By default it expects the style to be consistent, see [this](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004---unordered-list-style)
> 
> Since there is a `+` on line 66 it expects that all the lists follow that.

ah ok, then it's nice!

pickfire:
We only see what's bad, is there something that could be done instead of this pattern, if there is maybe we should show it.

simonsan:
> We only see what's bad, is there something that could be done instead of this pattern, if there is maybe we should show it.

As this anti-pattern is focussed on "using a method that catches unexpected problems (implementation bugs) for handling an expected problem, such as a missing file.", Imho `use good error handling` is the answer to this? What would you say?

pickfire:
I think that is an option but at least not so easy to pickup, but at least we have some suggestions. I am not aware of any straightforward switch.

anti_patterns/catch_panic.md
```diff
@@ -0,0 +1,79 @@
...
+In rust, there are two ways an operation can fail: An expected problem, like a
+file not being found, or a HTTP request timing out. Or, an unexpected problem,
+such as an index being out of bounds, or an assert!() failing. These are
+unexpected because they are bugs that should not happen. It would not make sense
+to handle an error for QuickSort returning an incorrectly unsorted array, as
+this should not happen.
+
+## Advantages
+
+There are scenarios where using `panic::catch_unwind` is the correct choice, e.g. a
```
jhwgh1968, line 41:
```suggestion
When code is split between multiple threads, or wants to create an isolation boundary, `panic::catch_unwind` is the correct choice. For example, a
```

pickfire: I don't understand "wants to create an isolation boundary". But good to have solid examples.

jhwgh1968: I thought the phrase "isolation boundary" appeared in the nomicon, but looking for it again, I misremembered. I was thinking of something like exception safety [similar to this example](https://doc.rust-lang.org/nomicon/exception-safety.html#vecpush_all), but guarding against "untrusted" code rather than "unsafe" code.

In addition to perhaps threading, that was what I imagined the Web Server example was doing.

PR #115, opened 2021-01-01
add Continuous Integration

MarcoIeni:
In this way we check the book builds on each PR.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
```
simonsan, line 7: Do we just want to do it on master or every branch in the repository?

MarcoIeni: From what I've understood, if you don't specify the branch, it will run the CI twice when you commit on the PRs.
See https://github.com/dandavison/delta/pull/467

simonsan: ah ok, well then lgtm.

simonsan (approved):

PR #124, opened 2021-01-02
Add doc tests to CI

### LINKED ISSUE #49: rustdoc --test the examples

simonsan:
Add doc tests to CI an ignore some already existing examples

Fixes #49

### DIFF
--- a/anti_patterns/deref.md
+++ b/anti_patterns/deref.md
@@ -26,11 +26,16 @@ public static void main(String[] args) {
 
 We can use the deref polymorphism anti-pattern to do so:
 
-```rust
+```rust,ignore
+use std::ops::Deref;
+
 struct Foo {}
 
 impl Foo {
-    fn m(&self) { ... }
+    fn m(&self) { 
+        //.. 
+    }
+
 }
 
 struct Bar {
@@ -44,7 +49,7 @@ impl Deref for Bar {
     }
 }
 
-fn main() {
+fn baz() {
     let b = Bar { Foo {} };
     b.m();
 }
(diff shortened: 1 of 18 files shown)


simonsan:
We need to merge this PR first, because I needed to deactivate some branch protection rules because I renamed the `test` to `deploy-test`

anti_patterns/deref.md
```diff
@@ -44,7 +49,7 @@ impl Deref for Bar {
     }
 }
 
-fn main() {
+fn baz() {
```
MarcoIeni, line 52: why rename main?

MarcoIeni: I just tried and tests works if the function is named `main'.

simonsan: Might have misunderstood this comment wrong https://github.com/rust-unofficial/patterns/pull/23#discussion_r72037199 when skimming through the older PRs. I thought there shouldn't be `main( )` in examples

simonsan: And llogiq did it ins his initial PR as well:
https://github.com/rust-unofficial/patterns/pull/51/files

So I thought this was kind of a convention or so. Not totally sure though.

MarcoIeni: Just opened a random page of a random rust book. It's full of main functions:
https://rust-lang.github.io/async-book/01_getting_started/04_async_await_primer.html

having the main is good because you can quickly copy paste the code and run it.
If there are two functions it also shows which is the entrypoint of the program.

simonsan: Fixed it where applicable. I'm not sure if we should name functions that take a Mutex or a mutable variable main though? There are some cases still in here.

MarcoIeni (approved):

PR #134, opened 2021-01-02
Fix default idiom example

RileyMShea:
Before this change,

the example code doesn't run on the current rust stable release(1.49).

This is because `Path` inherently has a u8 that requires the sized trait which requires either statics or replacing Path with PathBuf.

After this change,

The example code will run "as-is" without warnings or errors.

### DIFF
--- a/idioms/default.md
+++ b/idioms/default.md
@@ -44,7 +44,7 @@ fn main() {
     let mut conf = MyConfiguration::default();
     // do something with conf here
     conf.check = true;
-    eprintln!("conf = {:#?}", conf);
+    println!("conf = {:#?}", conf);
 }
 ```
 
--- a/idioms/default.md
+++ b/idioms/default.md
@@ -19,9 +19,7 @@ different names, but there can only be one `Default` implementation per type.
 
 ## Example
 
-```rust,ignore
-// note that we can simply auto-derive Default here.
-#[derive(Default)]
+```rust
 use std::{path::PathBuf, time::Duration};
 
 // note that we can simply auto-derive Default here.
--- a/idioms/default.md
+++ b/idioms/default.md
@@ -22,11 +22,15 @@ different names, but there can only be one `Default` implementation per type.
 ```rust,ignore
 // note that we can simply auto-derive Default here.
 #[derive(Default)]
+use std::{path::PathBuf, time::Duration};
+
+// note that we can simply auto-derive Default here.
+#[derive(Default, Debug)]
 struct MyConfiguration {
     // Option defaults to None
-    output: Option<Path>,
-    // Vecs default to empty vector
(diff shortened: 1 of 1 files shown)


idioms/default.md
```diff
@@ -22,11 +22,15 @@ different names, but there can only be one `Default` implementation per type.
 ```rust,ignore
 // note that we can simply auto-derive Default here.
 #[derive(Default)]
+use std::{path::PathBuf, time::Duration};
```
MarcoIeni, line 25: remove the derive statement above (line 24) which is an error.
Than you can remove the ignore on line 22, since the test pass now! :D

idioms/default.md
```diff
@@ -22,11 +22,15 @@ different names, but there can only be one `Default` implementation per type.
...
+
+// note that we can simply auto-derive Default here.
+#[derive(Default, Debug)]
 struct MyConfiguration {
     // Option defaults to None
-    output: Option<Path>,
-    // Vecs default to empty vector
-    search_path: Vec<Path>,
+    output: Option<PathBuf>,
+    // // Vecs default to empty vector
```
MarcoIeni, line 32:
```suggestion
    // Vecs default to empty vector
```

why double `//` ?

idioms/default.md
MarcoIeni: Also you should remove comment on line 23 since it's duplicated

idioms/default.md
```diff
@@ -41,6 +45,8 @@ fn main() {
     // construct a new instance with default values
     let mut conf = MyConfiguration::default();
     // do somthing with conf here
+    conf.check = true;
+    eprintln!("conf = {:#?}", conf);
```
MarcoIeni, line 49:
```suggestion
    println!("conf = {:#?}", conf);
```

why `eprintln`?

idioms/default.md
```diff
@@ -41,6 +45,8 @@ fn main() {
     // construct a new instance with default values
     let mut conf = MyConfiguration::default();
     // do somthing with conf here
```
MarcoIeni, line 47:
```suggestion
    // do something with conf here
```

idioms/default.md
```diff
@@ -22,11 +22,15 @@ different names, but there can only be one `Default` implementation per type.
 ```rust,ignore
 // note that we can simply auto-derive Default here.
 #[derive(Default)]
```
MarcoIeni, line 24:
```suggestion
```rust
```

MarcoIeni: line 23 was a duplicated comment,
line 24 was an error I think

MarcoIeni: I removed the ignore on line 22, since the test passes now! :D

simonsan:
Thank you! :-)

PR #135, opened 2021-01-02
Updated extensibility to discuss non_exhaustive

### LINKED ISSUE #132: Should #[non_exhaustive] be advised over private members for extensibility?
`#[non_exhaustive]` attribute on Rust [reference](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).

I believe `#[non_exhaustive]` emerged as a better solution to the same problem and private members are an artifact from the rime before it. Still it seems there are a few differences.

From what I can understand, the attribute has three primary differences compared to private members:
* It is still exhaustive inside original crate. There can be cases where this is undesirable (but I struggle to imagine when) - in which case private members still can be useful.
* It can be applied to unit structs/variants. Impossible to achieve with private members (have to convert it to tuple/struct and tuple/struct variant respectively).
* It can be applied to enums. A different approach is required to achieve similar effect without the attribute - a variant with never type (empty enum or `!`). It is not mentioned in the book.

Overall I consider the new approach as much superior (less boilerplate, better consistency), the old one should be deprecated and used only when very specific semantics is required (e. g. module-level non-exhaustiveness vs crate-level).

rcoh:
A first cut at discussing how `#[non_exhaustive]` should be used.

Fixes #132


idioms/priv-extend.md
```diff
@@ -1,35 +1,52 @@
-# Privacy for extensibility
+# `#[non_exhaustive]` for extensibility
 
 ## Description
 
-Use a private field to ensure that a struct is extensible without breaking
+Use `#[non_exhaustive]` on `struct`, `enum` and `enum` variant definitions to ensure that a struct is extensible without breaking
```
MarcoIeni, line 5: enum and enum?

idioms/priv-extend.md
```diff
@@ -1,35 +1,52 @@
...
 stability guarantees.
 
-
+For extensive documentation on all the places `#[non_exhaustive]` can be used, see [the docs](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).
```
MarcoIeni, line 8:
```suggestion
For extensive documentation on all the places `#[non_exhaustive]` can be used, see [the docs](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).

```

idioms/priv-extend.md
rcoh: "Enum variant", maybe it should be hyphenated?

MarcoIeni: sounds good

rcoh: Updated to match how it's described in https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute

simonsan (dismissed):
I miss the discussion when to use `privacy` and when to use `#[non_exhaustive]`. There are different use cases aren't they?

We should compare them in here and give a small direction for people to let them decide when to use what. 


I like that read as well, to understand the background, a `# See also` section might be good to add.
```
# See also
  - https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md
```

And then the `# Advantages, # Disadvantages` section could state general arguments for/against(?) field extensibility. While the `# Discussion` point can discuss when to use which of them both.

pickfire:
I only see that the advantages being mentioned but not the disadvantages. Most of the time I see this as unnecessary because it prevents the compiler from doing its job to check that the developer handle all the fields. This could be one example, in most cases having a breaking change by adding a new field to an enum is better than using `#[non_exhaustive]` as the user may not even notice that there is a new field if they want to handle all fields.

rcoh:
I've fleshed out the discussion some more although I think there's still a bit more to do

idioms/priv-extend.md
```diff
@@ -43,10 +49,27 @@ fn main(s: a::S) {
...
 
+`#[non_exhaustive]` only works across crate boundaries. Within a crate, the private field method may be used:
+
+For `struct`s, an alternative approach exists: By adding a private field to a struct, the struct cannot be instantiated or matched against.
+
+```rust
+pub struct S {
+   pub a: i32,
+   // Because `b` is private, you cannot match on `S` without using `..` and `S` cannot be directly instantiated
+   b: ()
```
pickfire, line 60:
```suggestion
    pub a: i32,
    // Because `b` is private, you cannot match on `S` without using `..` and `S` cannot be directly instantiated
    b: ()
```

simonsan:
@pickfire @MarcoIeni Want to give it a read as well?

@rcoh What did you want to add to it still? Or is it fine from your side now?

rcoh:
good from my side

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
-stability guarantees.
+A small set of scenarios exist where a library author may want to add public
+fields to a public struct or new variants to an enum without breaking backwards
+compatibility. Rust offers two solutions:
+
+- Use `#[non_exhaustive]` on `struct`s, `enum`s, and `enum` variants.
+For extensive documentation on all the places `#[non_exhaustive]` can be used,
+see [the docs](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).
+
+- For `struct`s only, you may add a private field to struct
```
MarcoIeni, line 13: That's not entirely true.
If before you have:
```rust
struct Point {
    pub x: u32,
    pub y: u32
}
```

and than you do:
```rust
struct Point {
    pub x: u32,
    pub y: u32,
    z: u32
}
```

This is not backward compatible, because previous code like `let a = Point {x:1, y:2}` won't work anymore

MarcoIeni: You can add something like "if other private fields already existed" to the sentence in order to fix this.
In fact if other private fields existed, you were not already allowed to create a point like that.

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
+## Warning
+
+Use this deliberately and with caution: Incrementing the major version when adding
+fields or variants is often a better option. `#[non_exhaustive]` may be appropriate
+in scenarios where you're modeling an external resource that may change out-of-sync
+with your library, but is not a general purpose tool.
+
+`#[non_exhaustive]` forces clients to handle the "Something else" case; there is
+rarely a sensible action to take in this scenario. This leads to awkward code and
+code paths that only executed in extremely rare circumstances.
```
MarcoIeni, line 24:
```suggestion
code paths that are only executed in extremely rare circumstances.
```

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
 
 ## Example
 
 ```rust,ignore
```
MarcoIeni, line 28: can we remove this `ignore`?

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
+
+        // The wildcard match is required because more variants may be
+        // added in the future
+        _ => println1("it's a new variant");
+    }
 }
+```
+
+`#[non_exhaustive]` only works across crate boundaries. Within a crate, the
+private field method may be used:
```
MarcoIeni, line 67: The trailing `:` here is weird

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
+        // .. required because this variant is non-exhaustive as well
+        a::AdmitMoreVariants::VariantC { a, .. } => println!("it's a c");
+
+        // The wildcard match is required because more variants may be
+        // added in the future
+        _ => println1("it's a new variant");
+    }
 }
+```
+
```
MarcoIeni, line 65: The example finishes here. We should move the rest of the section to another section or create a new section.

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
-match expressions. A private variant would force clients to have a `_` wildcard
-pattern. A common way to implement this instead is using the [`#[non_exhaustive]`](<https://doc.rust-lang.org/reference/attributes/type_system.html>)
-attribute.
+On `struct`s `#[non_exhaustive]` allows adding additional fields in a backwards
+compatible way. It will also prevent clients from using the struct constructor,
+even if all the fields are public. This may be helpful, but it's worth considering
+if you _want_ an additional field to be found by clients as a compiler error rather
+than something that may be silently undiscovered.
+
+`#[non_exhaustive]` when applied to `enum`s forces clients to handle a wildcard variant.
```
MarcoIeni, line 89:
```suggestion
`#[non_exhaustive]`, when applied to `enum`s, forces clients to handle a wildcard variant.
```

idioms/priv-extend.md
```diff
@@ -1,46 +1,103 @@
...
+
```
MarcoIeni, line 90:
```suggestion
```

Same thing is written in "disadvantages"

idioms/priv-extend.md
simonsan:
```suggestion
```rust
```

Indeed, it's a simple enough example that we should keep working.

idioms/priv-extend.md
simonsan:
```suggestion
private field method may be used.
```

simonsan:
@rcoh I think only the last review comment is left and fixing the doc test and then we're good to go! :)

HKalbasi:
Seems @rcoh lost interest on this. Can a team member finish it? It's a pity that the old article is online while this is almost ready.

simonsan:
> 
> 
> Seems @rcoh lost interest on this. Can a team member finish it? It's a pity that the old article is online while this is almost ready.

Should be fine now, @MarcoIeni @pickfire might want to read over it once more so we can merge it.

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
-# Privacy for extensibility
+# `#[non_exhaustive]` & private fields for extensibility
```
MarcoIeni, line 1:
```suggestion
# `#[non_exhaustive]` and private fields for extensibility
```
maybe too nitpick

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
...
 
 ## Description
 
-Use a private field to ensure that a struct is extensible without breaking
-stability guarantees.
+A small set of scenarios exist where a library author may want to add public
+fields to a public struct or new variants to an enum without breaking backwards
+compatibility. Rust offers two solutions:
```
MarcoIeni, line 7:
```suggestion
compatibility.
Rust offers two solutions to this problem:
```

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
...
+
+- Use `#[non_exhaustive]` on `struct`s, `enum`s, and `enum` variants.
+For extensive documentation on all the places `#[non_exhaustive]` can be used,
+see [the docs](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).
```
MarcoIeni, line 11:
```suggestion
  For extensive documentation on all the places `#[non_exhaustive]` can be used,
  see [the docs](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).
```

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
...
+
+- You may add a private field to a struct to prevent it from being directly
+instantiated or matched against
```
MarcoIeni, line 14:
```suggestion
  instantiated or matched against
```

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
...
+
+## Warning
+
+Use this deliberately and with caution: Incrementing the major version when adding
```
MarcoIeni, line 18:
```suggestion
Use this deliberately and with caution: incrementing the major version when adding
```

idioms/priv-extend.md
```diff
@@ -1,31 +1,72 @@
...
+fields or variants is often a better option. `#[non_exhaustive]` may be appropriate
+in scenarios where you're modeling an external resource that may change out-of-sync
+with your library, but is not a general purpose tool.
+
+`#[non_exhaustive]` forces clients to handle the "Something else" case; there is
```
MarcoIeni, line 23:
```suggestion
In fact, `#[non_exhaustive]` forces clients to handle the "Something else" case; there is
```

idioms/priv-extend.md
MarcoIeni, line 14: I would remove

MarcoIeni: Actually I would remove this point completely, since it is discussed as an alternative below

idioms/priv-extend.md
```diff
@@ -39,8 +80,37 @@ The downside of this approach is that you might need to add an otherwise unneede
...
+    pub a: i32,
+    // Because `b` is private, you cannot match on `S` without using `..` and `S`
+    //  cannot be directly instantiated or matched against
+    _b: ()
+}
+```
+
+## Discussion
+
+On `struct`s `#[non_exhaustive]` allows adding additional fields in a backwards
```
MarcoIeni, line 94:
```suggestion
On `struct`s, `#[non_exhaustive]` allows adding additional fields in a backwards
```

idioms/priv-extend.md
```diff
@@ -39,8 +80,37 @@ The downside of this approach is that you might need to add an otherwise unneede
...
+compatible way. It will also prevent clients from using the struct constructor,
+even if all the fields are public. This may be helpful, but it's worth considering
+if you _want_ an additional field to be found by clients as a compiler error rather
+than something that may be silently undiscovered.
+
+Finally, #[non_exhaustive] can be applied to enum variants. A `#[non_exhaustive]`
```
MarcoIeni, line 100:
```suggestion
Finally, `#[non_exhaustive]` can be applied to enum variants. A `#[non_exhaustive]`
```

idioms/priv-extend.md
```diff
@@ -39,8 +80,37 @@ The downside of this approach is that you might need to add an otherwise unneede
...
+variant behaves in the same way as a `#[non_exhaustive]` struct.
+
```
MarcoIeni, line 102:
```suggestion
```
I think this is just a repetition and this was already clear.

If we decide to keep it, let's add " ` " around #[non_exhaustive]

idioms/priv-extend.md
```diff
@@ -39,8 +80,37 @@ The downside of this approach is that you might need to add an otherwise unneede
...
+### Disadvantages
+
+`#[non_exhaustive]` can make your code much less ergonomic to use, especially when
+forced to handle unknown enum variants. It should only be used when these sorts of
+evolutions are required **without** incrementing the major version.
+
+When `#[non_exhaustive]` is applied to `enum`s, it forces clients to handle a
+wildcard variant. If there is no sensible action to take in this case, this may
+lead to brittle code. If a client decides to `panic!()` in this scenario, it may
+have been better to expose this error at compile time.
```
MarcoIeni, line 112: the `Disadvantages` should be merged with the `Warning` one, since they are very similar.
We can open an issue about this after we merge this maybe.

simonsan (approved):

simonsan:
Thank you everyone. :)

PR #136, opened 2021-01-02
Update compose structs compile error example for NLL

brandondong:
The first example in https://rust-unofficial.github.io/patterns/patterns/compose-structs.html states that there will be a compile error but this is not true with non lexical lifetimes.

### DIFF
--- a/patterns/compose-structs.md
+++ b/patterns/compose-structs.md
@@ -31,10 +31,11 @@ fn foo(a: &mut A) -> &u32 { &a.f2 }
 fn bar(a: &mut A) -> u32 { a.f1 + a.f3 }
 
 fn baz(a: &mut A) {
-    // x causes a to be borrowed for the rest of the function.
+    // The later usage of x causes a to be borrowed for the rest of the function.
     let x = foo(a);
     // Borrow check error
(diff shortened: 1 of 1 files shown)


simonsan:
Hey @brandondong thanks for pointing that out, as you're on it, could you please remove the `ignore` from the first example and make it pass the test? So we continuously improve test coverage of the examples.

patterns/compose-structs.md
```diff
@@ -31,10 +31,11 @@ fn foo(a: &mut A) -> &u32 { &a.f2 }
 fn bar(a: &mut A) -> u32 { a.f1 + a.f3 }
 
 fn baz(a: &mut A) {
-    // x causes a to be borrowed for the rest of the function.
+    // The later usage of x causes a to be borrowed for the rest of the function.
     let x = foo(a);
-    // Borrow check error
-    let y = bar(a); //~ ERROR: cannot borrow `*a` as mutable more than once at a time
+    // Borrow checker error:
+    // let y = bar(a); ~ ERROR: cannot borrow `*a` as mutable more than once at a time
```
simonsan, line 37:
```suggestion
    // let y = bar(a); // ~ ERROR: cannot borrow `*a` as mutable more than once at a time
```

simonsan:
Thank you! :-)

PR #137, opened 2021-01-02
Fix on-stack dispatch example

ovidiu:
* A `char` can't be used as a path. The correct `"-"` is already used in the "Disadvantages" example.
* Removed the `ignore` from the code block.


MarcoIeni:
Since you are working on this do you want to try to remove the `ignore` on line 12?
I tried to remove it and there are still some problems to fix!

ovidiu:
@MarcoIeni I think I've got it. Please have a look.

simonsan:
Thanks! ;-)

PR #143, opened 2021-01-03
fix internal links

MarcoIeni:
In this way they also work locally.


simonsan:
we might be also able to solve this if we rename README.md in those folders to index.md? not sure though need to try it out

simonsan:
That would also work it seems 🤔 @MarcoIeni Which one do you think is better. the `index.md` approach is github unfriendly because it doesn't automatically show the README on opening a folder. the initial version shows the readme but doesn't link on github 🤔

MarcoIeni:
Let's go with index.md
it is more clear because it converts to index.html

simonsan:
> 
> 
> Let's go with index.md
> it is more clear because it converts to index.html

Have you tested this change as well locally? For me it seems to work.

The disadvantage is all the open PRs that changes something in these files will need to be revises, but it's fine I guess

simonsan:
Ok, tested it again, should be fine. 👍🏽

PR #144, opened 2021-01-03
README: remove Table of Contents

MarcoIeni:
It is a duplicate of SUMMARY.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,46 +4,24 @@ An open source book about design patterns and idioms in the Rust programming
 language that you can read [here](https://rust-unofficial.github.io/patterns/).
 
 
-## Contents
-
-[Introduction](intro.md)
-
+## TODOs
 
 ### Idioms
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,46 +4,24 @@ An open source book about design patterns and idioms in the Rust programming
 language that you can read [here](https://rust-unofficial.github.io/patterns/).
 
 
-## Contents
```
simonsan, line 7: Maybe leaving this and linking to the SUMMARY.md could be nice?

MarcoIeni: mmm..Why the user might want to browse the book from github and not from the online version of the book?
It might be confusing for them

simonsan: Yeah, on first glance I don't have anything in mind as well. And we are always able to add that. So sure

simonsan (approved):

PR #146, opened 2021-01-03
Strategy

### LINKED ISSUE #120: Strategy design pattern
Working on https://medium.com/swlh/strategy-design-pattern-in-rust-5f5486cd294c?source=friends_link&sk=6871f4761dc6c4b7c0bc3ccefd22e408

fade2black:
Please review.

Closes #120


patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
+# Strategy (aka Policy)
+
+## Description
+
+The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) is a technique that enables separation of concern in object-oriented software design. It also allows to achieve decoupling software modules known as [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). The basic idea behind Strategy is given an algorithm, solving a particular problem, we define only skeleton of the algorithm at abstract level, but delegate specific algorithm’s implementation or some of its parts to another class or method. A client using algorithm may choose a specific implementation, while the general algorithm workflow remains the same. In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification. This is why we call it Dependency Inversion.
```
MarcoIeni, line 5:
```suggestion
The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) is a technique that enables separation of concerns in object-oriented software design. It also allows to achieve decoupling software modules known as [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). The basic idea behind Strategy is given an algorithm, solving a particular problem, we define only skeleton of the algorithm at abstract level, but delegate specific algorithm’s implementation or some of its parts to another class or method. A client using algorithm may choose a specific implementation, while the general algorithm workflow remains the same. In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification. This is why we call it Dependency Inversion.
```

MarcoIeni: It's difficult to do review and visualize diffs on single lines. Can you go to new line on every `.` or something like that?

fade2black: Do you mean typing 

 _This is sentence 1. 
 This is sentence 2.
 This is sentence 3._

Instead of 

_This is sentence 1. This is sentence 2. This is sentence 3._ ?

MarcoIeni: Yes

fade2black: Of course, if it does not violate formatting rules.

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+// This trait must be implemented by different formats
+pub trait Formatter {
+  fn run(&self, report: &Report);
+}
+
+// STRATEGIES:
+
+// Implementation for Json format
+mod vendor1 {
+  use super::*;
```
pickfire, line 23: Can this be formatted to 4 spaces?

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+```rust
+fn main() {
+  let val = Some("Rust");
+  let len_strategy = |s: &str| s.len();
+  let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
+
+  println!("len: {}", val.map(len_strategy).unwrap());
+  println!("first bite: {}", val.map(first_byte_strategy).unwrap());
+}
+```
```
pickfire, line 146: Maybe we could mention `serde-json` as an example? Like one could just swap out `serde-json` with `toml` easily.

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+impl Adder{
+  pub fn add<F>(x: u8, y: u8, f: F) -> u8
+      where F: Fn(u8, u8) -> u8 {
+      f(x,y)
+  }
+}
+
+fn main() {
+  let arith_adder = |x, y| x + y;
+  let bool_adder = |x, y| if x==1 || y==1 {1} else {0};
```
pickfire, line 126: I believe this needs to be formatted.

pickfire:
There is multiple way to do this, one is multiple files/modules, one is using feature flags, one is multiple crates. Maybe we should mention these methods on how to provide different strategies?

By the way, are you running gofmt rather than rustfmt here? It would be nice to just rust the rust code through rustfmt and add it there. It would also be good to have some use cases/test cases for the examples, could be done using `assert_eq!`.

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Motivation
+
+Suppose we are working on a project whose purpose is periodically read big data, [wrangle](https://en.wikipedia.org/wiki/Data_wrangling) it, and finally store it in a database as key-value pairs. After the data are stored in database, we need to provide conumers with APIs to query data. However, different consumers may want data in different formats (json, txt, xml, etc), even in new one we have never seen so far. But we don't know in advance in what format consumers would query data. So, every time a request for a new format implementation comes, we have to rebuild our software. We'd also like to change our code as little as possible. Ideally, the wrangling part of our softaware shouldn't change because it has nothing to with data serialization. This is a one scenario where we could apply the Strategy pattern to design our software. We could aslo publish the wrangling part as a library leaving the serialization implementation to library users. The following example explains the idea.
```
pickfire, line 9: This example is quite huge, maybe we should use a simpler one but show `serde-json` as a use case on how the formatting is done using either pretty formatter or compact formatter for a bigger example?

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+
+```rust
+// This trait must be implemented by different formats
+pub trait Formatter {
+  fn run(&self, report: &Report);
+}
+
+// STRATEGIES:
+
+// Implementation for Json format
```
pickfire, line 21: This is an example that have no focus on performance, we should state that the implementation below is inefficient and should not be used.

patterns/strategy.md
simonsan, line 9:
```suggestion
Suppose we are working on a project whose purpose is periodically read big data, [wrangle](https://en.wikipedia.org/wiki/Data_wrangling) it, and finally store it in a database as key-value pairs.
After the data are stored in database, we need to provide conumers with APIs to query data.
However, different consumers may want data in different formats (json, txt, xml, etc), even in new one we have never seen so far. But we don't know in advance in what format consumers would query data.
So, every time a request for a new format implementation comes, we have to rebuild our software.
We'd also like to change our code as little as possible.
Ideally, the wrangling part of our softaware shouldn't change because it has nothing to with data serialization.
This is a one scenario where we could apply the Strategy pattern to design our software.
We could aslo publish the wrangling part as a library leaving the serialization implementation to library users. The following example explains the idea.
```

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+
+  println!("Plain text format:");
+  text_report.format();
+}
+
+```
+
+## Advantages
+
+Separation of concerns. In the previous example, Report does not know anything about specific implementations of `Vendor1::Json` and `Vendor2::Text`, whereas the output implementations does not care about how data is preprocessed, stored, and fetched. The only thing they have to know is context and and a specific trait and method to implement, i.e., `Generator` and `run`.
```
simonsan, line 105:
```suggestion
Separation of concerns. In the previous example, Report does not know anything about specific implementations of `Vendor1::Json` and `Vendor2::Text`, whereas the output implementations does not care about how data is preprocessed, stored, and fetched. 
The only thing they have to know is context and and a specific trait and method to implement, i.e., `Generator` and `run`. 
```

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+
+```
+
+## Advantages
+
+Separation of concerns. In the previous example, Report does not know anything about specific implementations of `Vendor1::Json` and `Vendor2::Text`, whereas the output implementations does not care about how data is preprocessed, stored, and fetched. The only thing they have to know is context and and a specific trait and method to implement, i.e., `Generator` and `run`. 
+
+## Disadvantages
+
+For each strategy there must me implemented at least one module, so number of modules increases with number of strategies. If there are many strategies to choose from then users have to know how strategies differ from one another.
```
simonsan, line 109:
```suggestion
For each strategy there must me implemented at least one module, so number of modules increases with number of strategies.
If there are many strategies to choose from then users have to know how strategies differ from one another.
```

patterns/strategy.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Discussion
+
+In the previous example all strategies in a single file. Typically in Rust, each strategy should be implemented in a separate module file. However, we don't need to use traits in order to design this pattern in Rust. The following toy example demonstates the idea of the Strategy pattern using Rust closures
```
simonsan, line 113:
```suggestion
In the previous example all strategies in a single file.
Typically in Rust, each strategy should be implemented in a separate module file.
However, we don't need to use traits in order to design this pattern in Rust.

The following toy example demonstrates the idea of the Strategy pattern using Rust `closures`:
```

simonsan (changes requested):

patterns/strategy.md
fade2black: Yes, of course.

patterns/strategy.md
fade2black: Good idead.

patterns/strategy.md
fade2black: Do you mean spaces? Or one liner `if x==1...`

patterns/strategy.md
fade2black: I agree.

patterns/strategy.md
fade2black: Do you mean dynamic dispatch against static?

simonsan:
@fade2black I did a bit of layouting before to actually track the changes better, don't be surprised when you pull the changes from your repo. ;-) Cheers.

patterns/strategy.md
fade2black: BTW, I don't use `gofmt`. I use `rustfmt 1.4.18-stable`. As for indentation, I use VS Code, and I set number of spaces to 2. `rust fmt` fixes it to four.

fade2black:
> There is multiple way to do this, one is multiple files/modules, one is using feature flags, one is multiple crates. Maybe we should mention these methods on how to provide different strategies?

@pickfire I agree. Actually, I mentioned it in the Discussion section in a single sentence: "Typically in Rust, each strategy should be implemented in a separate module file.". I'll add more on that.

patterns/strategy.md
pickfire: Both spaces and that.

patterns/strategy.md
pickfire: No, I meant the way it uses `print!`.

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
+# Strategy (aka Policy)
+
+## Description
+
+The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) is a technique that enables separation of concerns in object-oriented software design.
+It also allows to achieve decoupling software modules known as [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
+The basic idea behind Strategy is given an algorithm, solving a particular problem, we define only skeleton of the algorithm at abstract level, but delegate specific algorithm’s implementation or some of its parts to another class or method.
+A client using algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
```
pickfire, line 8: We need spacing in between paragraph, one large chuck of text is hard to read.

patterns/strategy.md
fade2black: @pickfire I have removed `prints`. Added `assert_eq!`s.

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification.
+This is why we call it Dependency Inversion.
+
+## Motivation
+
+Imagine we are working on a project that generates reports every month. 
+We need the reports to be generated in different formats, e.g., in `JSON` or `Plain Text` formats.
```
pickfire, line 15:
```suggestion
We need the reports to be generated in different formats (strategies), e.g., in `JSON` or `Plain Text` formats. 
```

We should relate it back to "strategy".

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+
+Imagine we are working on a project that generates reports every month. 
+We need the reports to be generated in different formats, e.g., in `JSON` or `Plain Text` formats. 
+But thing very and we don't know what kind of requiremnt we may get in the future. 
+For example, we may need to generate our report in a completly new format, 
+or just modify the existing format. 
+So, in order to make chages to our existing code base as much as possible 
+while keeping our solution adoptable to such changes, 
+we might exploit Strategy design pattern. 
+Rust has powerful trait system which allows to implement this pattern.
```
pickfire, line 22: Is this line necessary?

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+trait Formatter {
+    fn run(&self, context: &Context) -> Result;
+}
+
+struct Report;
+
+impl Report {
+    fn generate<T: Formatter>(g: T) -> Result {
+        //perform here backend operations which should not bother caller...
+        //fetch data from database
```
pickfire, line 49:
```suggestion
        // backend operations...
```

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+        let values = vec![1, 2];
+        // generate report
+        g.run(&Context { keys, values })
+    }
+}
+
+struct Text;
+impl Formatter for Text {
+    fn run(&self, context: &Context) -> Result {
+        let mut s = String::new();
```
pickfire, line 60: This part should take `&mut String` rather than creating string every time it runs.

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+In this example our invariants (or abstractions) are `Context`, `Formatter`, and `Report`. 
+`Text` and `Json` are our strategy structs. 
+These strategies have to implement `Formatter` trait.
+
+```rust
+use std::fmt::Write as FmtWrite;
+use std::{error, result};
+type Result = result::Result<String, Box<dyn error::Error>>;
+
+struct Context {
```
pickfire, line 35: I think we should not use `Context` as an example. Too many places uses the word "context" especially in golang which may make people confuse what is "context". We should have a more specific word for this situation.

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+}
+
+struct Report;
+
+impl Report {
+    fn generate<T: Formatter>(g: T) -> Result {
+        //perform here backend operations which should not bother caller...
+        //fetch data from database
+        let keys = vec!["one".to_string(), "two".to_string()];
+        let values = vec![1, 2];
```
pickfire, line 51: Shouldn't we show a case with `HashMap` for this? If we use two vecs, it means that it could have different length.

patterns/strategy.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Disadvantages
+
+For each strategy there must me implemented at least one module, so number of modules increases with number of strategies.
+If there are many strategies to choose from then users have to know how strategies differ from one another.
+
+## Discussion
+
+In the previous example all strategies in a single file.
+Typically in Rust, each strategy should be implemented in a separate module file.
```
pickfire, line 107:
```suggestion
Ways of providing different strategies includes:

- All in one file (as shown in this example)
- Separated as modules, E.g. `formatter::json` module, `formatter::text` module
- Use compiler feature flags, E.g. `json` feature, `text` feature 
- Separated as crates, E.g. `json` crate, `text` crate
```

We may need to link each individual method, especially the use compiler feature flags.

patterns/strategy.md
fade2black: Shouldn't length of keys and length of values should be the equal? We assume each value corresponds to each key. I think we it is not that important whether we use `HashMap` or another collection. The main focus should be how we apply Strategy. After all, it is a toy example. I am also planning mention `serde` with use example.

patterns/strategy.md
fade2black: Do you mean `run(&self, context: &Context, s: &mut String) -> Result` ?

patterns/strategy.md
fade2black: After the second thought, using `HashMap` allows getting rid of `Context` and ugly `for` loop on vecs.

simonsan:
👍🏽 Nice work ;-)

fade2black:
@simonsan While I was looking through patterns on this repo, I noticed that people does not mention or refer to [SOLID](https://en.wikipedia.org/wiki/SOLID) principles at all. When designing application we usually think in terms of design principles, e.g, DRY, SOLID, etc, and only then we choose specific patterns to achieve our goals. I think it would be, in particular, useful and informative to create a section for design principles too. I mean, sometimes Strategy, Command, etc,...patterns do not mean much. But if I know, for example, that a specific pattern allows me to to stay DRY, then I consider it.

patterns/strategy.md
```diff
@@ -0,0 +1,156 @@
...
+In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification.
+This is why we call it Dependency Inversion.
+
+## Motivation
+
+Imagine we are working on a project that generates reports every month. 
+We need the reports to be generated in different formats (strategies), e.g., in `JSON` or `Plain Text` formats. 
+But things very and we don't know what kind of requirement we may get in the future. 
+For example, we may need to generate our report in a completly new format, 
+or just modify the existing format.
```
MarcoIeni, line 19:
```suggestion
or just modify the existing format.

```

patterns/strategy.md
```diff
@@ -0,0 +1,156 @@
+# Strategy (aka Policy)
+
+## Description
+
+The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) is a technique that enables separation of concerns in object-oriented software design.
+It also allows to achieve decoupling software modules known as [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
```
MarcoIeni, line 6:
```suggestion
It also allows to decouple software modules through [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
```
Do you prefer it this way?

patterns/strategy.md
```diff
@@ -0,0 +1,156 @@
...
+The basic idea behind Strategy is given an algorithm, solving a particular problem, we define only skeleton of the algorithm at abstract level, but delegate specific algorithm’s implementation or some of its parts to another class or method.
```
MarcoIeni, line 7:
```suggestion
The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem, we define only the skeleton of the algorithm at an abstract level and we delegate the specific algorithm’s implementation (or some of its parts) to another class or method.
```

patterns/strategy.md
```diff
@@ -0,0 +1,156 @@
...
+
+A client using algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
```
MarcoIeni, line 9:
```suggestion
In this way, a client using the algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem, we define only the skeleton of the algorithm at an abstract level and we delegate the specific algorithm’s implementation (or some of its parts) to another class or method.
+
+In this way, a client using the algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
+In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification.
+This is why we call it "Dependency Inversion".
+
+## Motivation
+
+Imagine we are working on a project that generates reports every month.
+We need the reports to be generated in different formats (strategies), e.g., in `JSON` or `Plain Text` formats.
```
MarcoIeni, line 16:
```suggestion
We need the reports to be generated in different formats (strategies), e.g. `JSON` or `Plain Text`. 
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+
+In this way, a client using the algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
+In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification.
+This is why we call it "Dependency Inversion".
+
+## Motivation
+
+Imagine we are working on a project that generates reports every month.
+We need the reports to be generated in different formats (strategies), e.g., in `JSON` or `Plain Text` formats. 
+But things very and we don't know what kind of requirement we may get in the future.
```
MarcoIeni, line 18:
```suggestion
But things vary over time and we don't know what kind of requirement we may get in the future.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+For example, we may need to generate our report in a completly new format,
+or just modify the existing format.
```
MarcoIeni, line 20:
```suggestion
or just modify one of the existing formats.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Example
+
+In this example our invariants (or abstractions) are `Context`, `Formatter`, and `Report`
+`Text` and `Json` are our strategy structs.
```
MarcoIeni, line 25:
```suggestion
In this example our invariants (or abstractions) are `Context`, `Formatter`, and `Report`,
while `Text` and `Json` are our strategy structs.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+These strategies have to implement `Formatter` trait.
```
MarcoIeni, line 26:
```suggestion
These strategies have to implement the `Formatter` trait.
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+```rust
+fn main() {
+    let val = Some("Rust");
+    let len_strategy = |s: &str| s.len();
+    let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
+
+    assert_eq!(4, val.map(len_strategy).unwrap());
+    assert_eq!(82, val.map(first_byte_strategy).unwrap());
+}
+
```
MarcoIeni, line 160:
```suggestion
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+            0
+        }
+    };
+    let custom_adder = |x, y| 2 * x + y;
+
+    assert_eq!(9, Adder::add(4, 5, arith_adder));
+    assert_eq!(0, Adder::add(0, 0, bool_adder));
+    assert_eq!(5, Adder::add(1, 3, custom_adder));
+}
+
```
MarcoIeni, line 146:
```suggestion
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+fn main() {
+    let s = Report::generate(Text).unwrap();
+    assert!(s.contains("one 1"));
+    assert!(s.contains("two 2"));
+
+    let s = Report::generate(Json).unwrap();
+    assert!(s.contains(r#"{"one":"1"}"#));
+    assert!(s.contains(r#"{"two":"2"}"#));
+}
+
```
MarcoIeni, line 89:
```suggestion
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+In this example our invariants (or abstractions) are `Context`, `Formatter`, and `Report`,
+while `Text` and `Json` are our strategy structs.
+These strategies have to implement the `Formatter` trait.
+
+```rust
+use std::collections::HashMap;
+use std::fmt::Write as FmtWrite;
+use std::{error, result};
+type Result = result::Result<String, Box<dyn error::Error>>;
```
MarcoIeni, line 32: I think that this is confusing. I would prefer to use `Result<String, Box<dyn error::Error>>` everywhere it is necessary, instead of creating a type alias.
What other reviewers think?

MarcoIeni: Maybe we can call it something like `AnyResult` and we can remove the `use std::{error, result};` line?

fade2black: I added `use std::{error, result};` to avoid compiler errors because I used `Result`. Now that your offer `AnyResult` we can remove `use std::{error, result};`.

fade2black: Other reviewers hasn't sent yet anything concerning the alias for `Result<String, Box<dyn error::Error>>`. If it is really confusing I can remove and use `Result<String, Box<dyn error::Error>>`.

MarcoIeni: yes, sorry, only `Result` is in the prelude apparently

patterns/strategy.md
```diff
@@ -0,0 +1,160 @@
...
+        for (key, val) in data {
+            write!(&mut s, "{} {}\n", key, val)?;
+        }
+        Ok(s)
+    }
+}
+
+struct Json;
+impl Formatter for Json {
+    fn run(&self, data: &Data) -> Result<String, Box<dyn error::Error>> {
```
MarcoIeni, line 65:
```suggestion
use std::error::Error;
type Data = HashMap<String, u32>;

trait Formatter {
    fn run(&self, data: &Data) -> Result<String, Box<dyn Error>>;
}

struct Report;

impl Report {
    fn generate<T: Formatter>(g: T) -> Result<String, Box<dyn Error>> {
        // backend operations...
        let mut data = HashMap::new();
        data.insert("one".to_string(), 1);
        data.insert("two".to_string(), 2);
        // generate report
        g.run(&data)
    }
}

struct Text;
impl Formatter for Text {
    fn run(&self, data: &Data) -> Result<String, Box<dyn Error>> {
        let mut s = String::new();

        for (key, val) in data {
            write!(&mut s, "{} {}\n", key, val)?;
        }
        Ok(s)
    }
}

struct Json;
impl Formatter for Json {
    fn run(&self, data: &Data) -> Result<String, Box<dyn Error>> {
```

patterns/strategy.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Example
+
+In this example our invariants (or abstractions) are `Context`, `Formatter`, and `Report`,
+while `Text` and `Json` are our strategy structs.
+These strategies have to implement the `Formatter` trait.
+
+```rust
+use std::collections::HashMap;
+use std::fmt::Write as FmtWrite;
```
MarcoIeni, line 30:
```suggestion
use std::fmt::Write;
```

This has no impact, why do you do it?

patterns/strategy.md
MarcoIeni: why? If that's only for performance reason..come on, it's just an example! :)

patterns/strategy.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+struct Text;
+impl Formatter for Text {
+    fn run(&self, data: &Data) -> Result<String, Box<dyn Error>> {
+        let mut s = String::new();
+
+        for (key, val) in data {
+            write!(&mut s, "{} {}\n", key, val)?;
```
MarcoIeni, line 57:
```suggestion
            let entry = format!("{} {}\n", key, val);
            s.push_str(&entry);
```
You use `write!` that may fail, but if you use something like this you are sure that it doesn't fail.
So basically all the `Result`s can be removed.

patterns/strategy.md
fade2black: Probably leftover. :-)

patterns/strategy.md
MarcoIeni: This `Text`, in particular, could be written really elegantly with `map` I believe, but that's not required, it's just a thought I had.

MarcoIeni:
I am rewriting this here because the conversation was resolved.
fade2black, what do you think about removing all the `Result` in the example by substituting `write!` with something like `format!` and `push_str`?

fade2black:
@MarcoIeni Agree. I'm removing `Result`s.

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+
+    let s = Report::generate(Json).unwrap();
+    assert!(s.contains(r#"{"one":"1"}"#));
+    assert!(s.contains(r#"{"two":"2"}"#));
+}
+```
+
+## Advantages
+
+Separation of concerns. In the previous example, Report does not know anything about specific
```
MarcoIeni, line 93:
```suggestion
Separation of concerns. In the previous example, `Report` does not know anything about specific
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+implementations of `Json` and `Text`, whereas the output implementations does not care about how data is
+preprocessed, stored, and fetched.
+The only thing they have to know is context and and a specific trait and method to implement,
```
MarcoIeni, line 96:
```suggestion
The only thing they have to know is context and a specific trait and method to implement,
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+i.e,`Formatter` and `run`.
+
+## Disadvantages
+
+For each strategy there must me implemented at least one module, so number of modules
```
MarcoIeni, line 101:
```suggestion
For each strategy there must be implemented at least one module, so number of modules
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+increases with number of strategies.
+If there are many strategies to choose from then users have to know how strategies differ
+from one another.
+
+## Discussion
+
+In the previous example all strategies in a single file.
```
MarcoIeni, line 108:
```suggestion
In the previous example all strategies are implemented in a single file.
```

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+In fact, Rust already uses this idea for `Options`'s `map` method
+
+```rust
+fn main() {
+    let val = Some("Rust");
+    let len_strategy = |s: &str| s.len();
+    let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
+
+    assert_eq!(4, val.map(len_strategy).unwrap());
+    assert_eq!(82, val.map(first_byte_strategy).unwrap());
```
MarcoIeni, line 159:
```suggestion
    let val = Some("Rust");

    let len_strategy = |s: &str| s.len();
    assert_eq!(4, val.map(len_strategy).unwrap());

    let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
    assert_eq!(82, val.map(first_byte_strategy).unwrap());
```

MarcoIeni: do you find it clearer like this?

MarcoIeni: (I just moved the instructions)

fade2black: 👍

patterns/strategy.md
```diff
@@ -0,0 +1,161 @@
...
+
+    let s = Report::generate(Json).unwrap();
+    assert!(s.contains(r#"{"one":"1"}"#));
+    assert!(s.contains(r#"{"two":"2"}"#));
+}
+```
+
+## Advantages
+
+Separation of concerns. In the previous example, `Report` does not know anything about specific
```
MarcoIeni, line 93:
```suggestion
The main advantage is separation of concerns. For example, in this case `Report` does not know anything about specific
```

fade2black:
@MarcoIeni Did you have in mind this 
```rust
data.iter().map(|(key, val)| format!("{} {}\n", key, val)).collect();
``` 
?

fade2black:
@MarcoIeni I use `return s` instead of simply typing `s`. `s` alone looks lost 😀

MarcoIeni:
> @MarcoIeni Did you have in mind this?
> 
> ```rust
> data.iter().map(|(key, val)| format!("{} {}\n", key, val)).collect();
> ```

Exactly! :heart: 

> @MarcoIeni I use `return s` instead of simply typing `s`. `s` alone looks lost

So let's remove both! :stuck_out_tongue_winking_eye: Suggestion incoming..

patterns/strategy.md
```diff
@@ -0,0 +1,160 @@
...
+}
+
+struct Text;
+impl Formatter for Text {
+    fn run(&self, data: &Data) -> String {
+        let s: String = data
+            .iter()
+            .map(|(key, val)| format!("{} {}\n", key, val))
+            .collect();
+        return s;
```
MarcoIeni, line 56:
```suggestion
        data
            .iter()
            .map(|(key, val)| format!("{} {}\n", key, val))
            .collect()
```

patterns/strategy.md
```diff
@@ -0,0 +1,159 @@
...
+        // generate report
+        g.run(&data)
+    }
+}
+
+struct Text;
+impl Formatter for Text {
+    fn run(&self, data: &Data) -> String {
+        data
+            .iter()
```
MarcoIeni, line 53:
```suggestion
        data.iter()
```
This is how rustfmt likes it

MarcoIeni (approved):

MarcoIeni:
I approved, but I don't merge this because maybe simonsan and pickfire want to review this as well :)

Anyway thank you so much for your contribution!!

MarcoIeni (approved):

fade2black:
@MarcoIeni Thank you too for your suggestions and collaboration!!!

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
+# Strategy (aka Policy)
+
+## Description
+
+The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) is a technique that enables separation of concerns in object-oriented software design.
```
MarcoIeni, line 5:
```suggestion
The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
is a technique that enables separation of concerns in object-oriented software design.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+It also allows to decouple software modules through [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
+The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem, we define only the skeleton of the algorithm at an abstract level and we delegate the specific algorithm’s implementation (or some of its parts) to another class or method.
```
MarcoIeni, line 7:
```suggestion
The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem,
we define only the skeleton of the algorithm at an abstract level and
we delegate the specific algorithm’s implementation (or some of its parts) to another class or method.
```

patterns/strategy.md
```diff
@@ -0,0 +1,158 @@
...
+
+In this way, a client using the algorithm may choose a specific implementation, while the general algorithm workflow remains the same.
+In other words, the abstract specification of the class does not depend on the specific implementation of the derived class, but specific implementation must adhere to the abstract specification.
```
MarcoIeni, line 10:
```suggestion
In other words, the abstract specification of the class does not depend on the specific implementation of the derived class,
but specific implementation must adhere to the abstract specification.
```

MarcoIeni:
I think it's time to merge this! If pickfire or anyone else want to propose changes they will be able to do a PR whenever they want! :)
I proposed some changes that should fix the CI. Once the CI passes I would like to merge this.

simonsan:
> 
> 
> I think it's time to merge this! If pickfire or anyone else want to propose changes they will be able to do a PR whenever they want! :)
> I proposed some changes that should fix the CI. Once the CI passes I would like to merge this.

Nothing for a `# See also` section? :(

MarcoIeni:
Maybe fade2black have some useful links to share!

fade2black:
@MarcoIeni I have added See also section. But I had to remove this link [Policy Based Design](https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design#Policy-based_design) from this section because the CI strangely complaints it is a dead link.

patterns/strategy.md
```diff
@@ -0,0 +1,168 @@
...
+    let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
+    assert_eq!(82, val.map(first_byte_strategy).unwrap());
+}
+```
+
+## See also
+
+[Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
+
+[Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)
```
MarcoIeni, line 168:
```suggestion
- [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
- [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)<!-- markdown-link-check-disable -->
- [Policy Based Design](https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design#Policy-based_design)<!-- markdown-link-check-enable -->
```

MarcoIeni: let's try if this works..

fade2black:
Passed.

MarcoIeni (dismissed):

fade2black:
@simonsan @MarcoIeni Serde is mentioned as an example.

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
+# Strategy (aka Policy)
+
+## Description
+
+The [Strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
+is a technique that enables separation of concerns in object-oriented software design.
```
pickfire, line 6: Is this really related to object-oriented software design? I can't see how it relates.

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
...
+It also allows to decouple software modules through [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
+The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem,
```
pickfire, line 8:
```suggestion

The basic idea behind the Strategy pattern is that, given an algorithm solving a particular problem,
```

A new line would be good to separate the basic idea.

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
...
+we define only the skeleton of the algorithm at an abstract level and
+we delegate the specific algorithm’s implementation (or some of its parts) to another class or method.
```
pickfire, line 10:
```suggestion
we separate the specific algorithm’s implementation into different parts.
```

I think using "another class or method" here may confuse people, in rust "class" is not a (strong) concept but it is borrowed from other languages so not everyone can understand it.

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
...
+type Data = HashMap<String, u32>;
+
+trait Formatter {
+    fn run(&self, data: &Data) -> String;
+}
+
+struct Report;
+
+impl Report {
+    fn generate<T: Formatter>(g: T) -> String {
```
pickfire, line 43: Ah, like I mentioned before. It's just an example. Since we are in rust patterns, I think we should show a good pattern since others will be learning from here. I still think taking in `&mut String` should be used here, at least others not knowing this will be more aware about it, it won't make it much complicated right?

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
...
+        data.insert("one".to_string(), 1);
+        data.insert("two".to_string(), 2);
+        // generate report
+        g.run(&data)
+    }
+}
+
+struct Text;
+impl Formatter for Text {
+    fn run(&self, data: &Data) -> String {
```
pickfire, line 55: Just my two cents, it would be better to change `run` to `format`, since it is named formatter, not really a pattern but run is very general.

patterns/strategy.md
fade2black: @pickfire We may discuss here on what OOD is for days and stil can't come to a consensus, can we?. Rust book ([chapter 17]( https://doc.rust-lang.org/book/ch17-01-what-is-oo.html)) writes:

> There is no consensus in the programming community about what features a language must have to be considered object oriented.
Nevertheless, in my opinion, if you are dealing with DATA + BEHAVIOUR then we traditionally call them "OBJECTS", and if we think in terms of "OBJECTS" then we simply call it object oriented.  

As for design patterns, they are simply a set of principles (or whatever you call them) to follow to make software design clear, easy to refactor, readable, etc.
The main point of this section is the Strategy pattern and not whether or not it relates to OOD or what OOD is.

patterns/strategy.md
```diff
@@ -0,0 +1,171 @@
...
+
+        for (key, val) in data {
+            if s.len() > 1 {
+                s.push(',');
+            }
+            entry = format!("{{\"{}\":\"{}\"}}", key, val);
+            s.push_str(&entry);
+        }
+        s.push(']');
+        return s;
```
pickfire, line 76:
```suggestion
        let mut s = String::from("[");
        let mut iter = data.into_iter();
                            
        if let Some((key, val)) = data.next() {
            let entry = format!(r#"{"{}":"{}"}"#, key, val);
            s.push_str(&entry);
            while let Some((key, val)) = data.next() {
                s.push(',');   
                let entry = format!(r#"{"{}":"{}"}"#, key, val);
                s.push_str(&entry);
            }
        }
        s.push(']');
        s
```
This should be more efficient since we unrolled the loop to remove the branch, ideally we want to make it branchless. Like the above example, it may be a bit longer but it is more efficient.

There is also another method to this, we could collect into `Vec<String>` and `slices::join` them, although it requires double allocation which could make it shorter but less efficient maybe.

patterns/strategy.md
pickfire: Okay, I guess I am fine with it after you explained, but I rarely see the docs mentioning about objects which make me a bit confused.

>  The main point of this section is the Strategy pattern and not whether or not it relates to OOD or what OOD is.

Yes, that is why I mention this. Because could be done in non OOD as well, since it could be separated from file level.

patterns/strategy.md
```diff
@@ -0,0 +1,174 @@
...
+In the previous example all strategies are implemented in a single file.
+Ways of providing different strategies includes:
+
+- All in one file (as shown in this example, similar to being separated as modules)
+- Separated as modules, E.g. `formatter::json` module, `formatter::text` module
+- Use compiler feature flags, E.g. `json` feature, `text` feature
+- Separated as crates, E.g. `json` crate, `text` crate
+
+Serde crate is a good example of the `Strategy` pattern in action. Serde allows [full customization](https://serde.rs/custom-serialization.html)
+of the serialization behavior by manually implementing `Serialize` and `Deserialize` traits for our type.
```
pickfire, line 120: Would be good to have an example for serde strategy.

```suggestion
Serde crate is a good example of the `Strategy` pattern in action. Serde allows [full customization](https://serde.rs/custom-serialization.html)
of the serialization behavior by manually implementing `Serialize` and `Deserialize` traits for our type.
For example, we could easily swap `serde_json` with `serde_cbor` since they expose a similar method.
Having this makes the helper crate `serde_transcode` much more useful and ergonomic.
```

patterns/strategy.md
```diff
@@ -0,0 +1,175 @@
...
+For example, we could easily swap `serde_json` with `serde_cbor` since they expose a similar method.
```
pickfire, line 120: Sorry, grammatical error I put just now.
```suggestion
For example, we could easily swap `serde_json` with `serde_cbor` since they expose similar methods.
```

patterns/strategy.md
```diff
@@ -0,0 +1,175 @@
...
+    let custom_adder = |x, y| 2 * x + y;
+
+    assert_eq!(9, Adder::add(4, 5, arith_adder));
+    assert_eq!(0, Adder::add(0, 0, bool_adder));
+    assert_eq!(5, Adder::add(1, 3, custom_adder));
+}
+
+```
+
+In fact, Rust already uses this idea for `Options`'s `map` method
```
pickfire, line 157:
```suggestion
In fact, Rust already uses this idea for `Options`'s `map` method:
```

patterns/strategy.md
```diff
@@ -0,0 +1,175 @@
...
+    let first_byte_strategy = |s: &str| s.bytes().next().unwrap();
+    assert_eq!(82, val.map(first_byte_strategy).unwrap());
+}
+```
+
+## See also
+
+- [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
+- [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)<!-- markdown-link-check-disable -->
+- [Policy Based Design](https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design#Policy-based_design)<!-- markdown-link-check-enable -->
```
pickfire, line 175: Weird, why do we need to disable this link check?

fade2black: @MarcoIeni ?

simonsan, line 175:
```suggestion
- [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)
- [Policy Based Design](https://en.wikipedia.org/wiki/Modern_C++_Design#Policy-based_design)
```

simonsan: Can we try that and see if it works, otherwise we revert and keep the disabled link check

simonsan: (Would be an upstream issue if it works like this)

simonsan:
Opened upstream issue:
https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/96

patterns/strategy.md
fade2black: I had in mind but didn't know how to do it in Rust. Good idiom: create iter, first next, and then iter til the end.

fade2black: not `data.next()`, but `iter.next()`

patterns/strategy.md
simonsan: @fade2black Seems like this is the last review comment that we could address before merging, any thoughts on it?

fade2black: @simonsan  I am on it.

fade2black: @simonsan @pickfire Fixed.

MarcoIeni: I think that taking the string as mut is not always the best approach.
The rust library is full of functions that returns strings as output.
Taking it as mut gives you flexibility, but it is more error prone and less readable. It depends on your priorities.
Anyway, I don't think it matters that much, so it's fine like this :)

MarcoIeni (approved):

patterns/strategy.md
fade2black: Whether it is `&mut String` or `String` as return, I don't care very much in this example. This is just an example.
Depending on use cases both `&mut String` and `String` as return would work fine. No need to waste time on [trivial things](https://en.wikipedia.org/wiki/Law_of_triviality).

simonsan:
Really nice, thank you! <3

PR #147, opened 2021-01-03
Add link to derive_more for newtypes

hobofan:
I think derive_more is one of the most helpful crates for mitigating the downsides of the newtype pattern.


MarcoIeni:
Yeah, it is also mentioned [here](https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html).

MarcoIeni (approved):

PR #117, opened 2021-01-01
Add introduction

simonsan:
Reviving the ghost PR #43 

Regarding `functional` and `refactoring` folders:

Something that we might add to the `issues`:
> I was thinking something similar to what exists in the `patterns` folder, probably with an altered template. So far my ideas are:
> 
>     * `functional`
>       
>       * currying
>       * higher-order functions
>       * collections
> 
>     * `refactoring`
>       
>       * refactoring from `unwrap` to `Result`
>       * refactoring from `enum` to `Option`

Some point may still stand? Need to check.

2021-01-01 15_00_03-Window-Screenshot

Fixes #4

PR #153, opened 2021-01-03
Use a problem as the intro for newtype pattern

pickfire:
A problem can highlight some of the use cases of newtype
pattern rather than discussing generally that rust is strongly
typed. Make the user understand the need of newtype.

I wonder if it should be in the description instead?

Continued https://github.com/rust-unofficial/patterns/pull/117

r? @simonsan

### DIFF
--- a/patterns/newtype.md
+++ b/patterns/newtype.md
@@ -1,8 +1,11 @@
 # Newtype
 
-Rust has strong static types. This can be very different than what you are used to if you are coming from a loosely-typed language. Don't worry, though. Once you get used to them, you'll find the types actually make your life easier. Why? Because you are making implicit assumptions explicit.
+What if in some cases we want a type to behave similar to another type but
+still different? What if we want to enforce some behavior at compile time
+which using only type aliases is not enough? For example, have a custom
+`Display` implementation for `String` for security reasons (password).
 
-A really convenient application of the Rust type system is the Newtype pattern.
+We could use the newtype pattern to provide type safety and encapsulation.
 
 ## Description
 
@@ -97,6 +100,7 @@ Here, `Bar` might be some public, generic type and `T1` and `T2` are some intern
 
 ## See also
 
+- [Advanced Types in the book](https://doc.rust-lang.org/book/ch19-04-advanced-types.html?highlight=newtype#using-the-newtype-pattern-for-type-safety-and-abstraction)
 - [Newtypes in the style guide](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html)
 - [Newtypes in Haskell](https://wiki.haskell.org/Newtype)
 - [Type aliases](https://doc.rust-lang.org/stable/book/ch19-04-advanced-types.html#creating-type-synonyms-with-type-aliases)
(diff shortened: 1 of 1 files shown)


patterns/newtype.md
```diff
@@ -97,6 +100,7 @@ Here, `Bar` might be some public, generic type and `T1` and `T2` are some intern
 
 ## See also
 
+- [Advanced Types in the book](https://doc.rust-lang.org/book/ch19-04-advanced-types.html?highlight=newtype#using-the-newtype-pattern-for-type-safety-and-abstraction)
 - [Newtypes in the style guide](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html)
```
pickfire, line 104: The style guide is outdated or not maintained IIRC.

patterns/newtype.md
```diff
@@ -1,8 +1,11 @@
 # Newtype
 
-Rust has strong static types. This can be very different than what you are used to if you are coming from a loosely-typed language. Don't worry, though. Once you get used to them, you'll find the types actually make your life easier. Why? Because you are making implicit assumptions explicit.
+What if in some cases we want a type to behave similar to another type but
+still different? What if we want to enforce some behavior at compile time
+which using only type aliases is not enough? For example, have a custom
```
simonsan, line 5:
```suggestion
which using only type aliases is not enough? For example, having a custom
```

patterns/newtype.md
simonsan: Sad, but true.

```suggestion
```

patterns/newtype.md
```diff
@@ -1,8 +1,11 @@
...
+`Display` implementation for `String` for security reasons (password).
```
simonsan, line 6:
```suggestion
What if in some cases we want a type to behave similar to another type or 
enforce some behaviour at compile time where using only type aliases would 
not be enough? 

For example, if you want to create a custom `Display` implementation for `String` 
due to security considerations (e.g. passwords).
```

patterns/newtype.md
```diff
@@ -1,8 +1,11 @@
...
 
-A really convenient application of the Rust type system is the Newtype pattern.
+We could use the newtype pattern to provide type safety and encapsulation.
```
simonsan, line 8:
```suggestion
For aforementioned and such cases we could use the `Newtype` pattern to provide __type safety and encapsulation__.
```

patterns/newtype.md
pickfire: The separation looks good but the word "you" does not look that good for me. Either way I just applied it.

patterns/newtype.md
pickfire:
```suggestion
For such cases we could use the `Newtype` pattern to provide __type safety__ and __encapsulation__.
```

aforementioned is a hard word, I think it is better to remove that.

simonsan (dismissed):

patterns/newtype.md
```diff
@@ -1,8 +1,13 @@
 # Newtype
 
-Rust has strong static types. This can be very different than what you are used to if you are coming from a loosely-typed language. Don't worry, though. Once you get used to them, you'll find the types actually make your life easier. Why? Because you are making implicit assumptions explicit.
+What if in some cases we want a type to behave similar to another type or 
+enforce some behaviour at compile time where using only type aliases would 
+not be enough? 
 
-A really convenient application of the Rust type system is the Newtype pattern.
+For example, if you want to create a custom `Display` implementation for `String`
```
simonsan, line 7:
```suggestion
For example, if we want to create a custom `Display` implementation for `String` 
```

simonsan (approved):

simonsan:
Nice, thank you! :-)

PR #175, opened 2021-01-06
markdownlint: fix md004

MarcoIeni:


simonsan:
We agreed yesterday in the initial PR that we would keep the hyphens instead of the `*`'s.

simonsan:
Not sure, are there any advantages of changing to `*`?

MarcoIeni:
> Not sure, are there any advantages of changing to `*`?

Of course not. This lint just checks that they are consistent across the same document. I will change to `-`.
Anyway, if we receive a PR that contains `*` we should tell (either we or the markdownlint) to switch to `-`? Or is it just the preferred style but it's optional?

simonsan:
> 
> 
> > Not sure, are there any advantages of changing to `*`?
> 
> Of course not. This lint just checks that they are consistent across the same document. I will change to `-`.
> Anyway, if we receive a PR that contains `*` we should tell (either we or the markdownlint) to switch to `-`? Or is it just the preferred style but it's optional?

Aaah, now I understand! :-D Thanks for clarifying!

Yeah, let's automate that for the people contributing that it's automatically switching to the corresponding style, or?

MarcoIeni:
> Yeah, let's automate that for the people contributing that it's automatically switching to the corresponding style, or?

The alternative is that every document has its own "bullet style": `-` or `*`.

For me it makes no difference. Both the approaches are fine, but it might be boring for a random contributor to switch from a style to the other after it wrote the entire page of the book.

Maybe ensuring the same style in each page is enough.

simonsan:
> 
> 
> > Yeah, let's automate that for the people contributing that it's automatically switching to the corresponding style, or?
> 
> The alternative is that every document has its own "bullet style": `-` or `*`.
> 
> For me it makes no difference. Both the approaches are fine, but it might be boring for a random contributor to switch from a style to the other after it wrote the entire page of the book.
> 
> Maybe ensuring the same style in each page is enough.

I think if it's automatically fixable by markdownlint that would be the best option. so we have a common style over all documents. In the end it's just hypens or stars true. But if we can easily fix it by letting run one command i think it should be not a big problem also for contributors that are used to another style.

README.md
```diff
@@ -7,35 +7,35 @@ language that you can read [here](https://rust-unofficial.github.io/patterns/).
...
-* TODO Matching all fields of a struct (back compat)
-* TODO wildcard matches
-* TODO taking an enum rather than having multiple functions
-* TODO `unwrap()`ing every `Result` instead of forwarding it
+- TODO thread + catch_panic for exceptions
+- TODO Clone to satisfy the borrow checker
+- TODO Matching all fields of a struct (back compat)
+- TODO wildcard matches
+- TODO taking an enum rather than having multiple functions
+- TODO `unwrap()`ing every `Result` instead of forwarding it
```
simonsan, line 38: Can you remove that and link to the umbrella issue with a small sentence stating that people will find all the content to add there?

simonsan: I can do that in another PR and then we merge this one after, maybe easier and cleaner

MarcoIeni: ok!

simonsan:
Is this LGTM?

MarcoIeni:
yes. Can you approve it?

simonsan (approved):

PR #178, opened 2021-01-06
do not allow tabs in code blocks

MarcoIeni:
See https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md010---hard-tabs

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,5 +1,3 @@
 ---
 MD004: false
-MD010:
-  code_blocks: false
 MD013: false
--- a/functional/index.md
+++ b/functional/index.md
@@ -7,7 +7,7 @@ Rust is an imperative language, but it follows many functional programming parad
 ```rust
 let mut sum = 0;
 for i in 1..11 {
-	sum += i;
+    sum += i;
 }
 println!("{}", sum);
 ```
--- a/idioms/pass-var-to-closure.md
+++ b/idioms/pass-var-to-closure.md
@@ -40,7 +40,7 @@ let num3 = Rc::new(3);
 let num2_cloned = num2.clone();
 let num3_borrowed = num3.as_ref();
 let closure = move || {
-	*num1 + *num2_cloned + *num3_borrowed;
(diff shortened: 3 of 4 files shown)


simonsan:
I think markdown lint should exclude/ignore codeblocks completely and we should enforce style inside the codeblock with a `cargo-fmt.toml`. I already opened an upstream issue for this.

MarcoIeni:
Yes, but in the meantime this is the best we can do.
Should we manually check if code contains tab in the meantime?

Also, what's the point of changing `markdownlint` default behavior by allowing tabs in code blocks? It's a sane default imho.

simonsan:
> 
> 
> Yes, but in the meantime this is the best we can do.
> Should we manually check if code contains tab in the meantime?
> 
> Also, what's the point of changing `markdownlint` default behavior by allowing tabs in code blocks? It's a sane default imho.

Naja, because I assume people write their code in their favoured IDE and adjust the style their with cargo fmt. We could state that in contributing as well, as soon as we have a `cargo-fmt.toml`. That they probably write their code in an `article.rs` file so that it works and gets checked by `rust-analyzer`/`RLS` and run `cargo fmt` then copy it over. Would let markdownlint mess with Rust code.

MarcoIeni:
Is `rustfmt` with default checks supported in mdbook at the moment? I.e. a similar code like `mdbook test` that checks for format.

simonsan:
> 
> 
> Is `rustfmt` with default checks supported in mdbook at the moment? I.e. a similar code like `mdbook test` that checks for format.

I opened an upstream issue for this for `mdbook fmt` which would be quite lovely, so i don't think so. But we could adjust the workflow given in contributing to let people know how we imagine it as mentioned above

simonsan:
Upstream issue is here
https://github.com/rust-lang/mdBook/issues/1422

MarcoIeni:
Yeah, this is awesome and what you have written is a very good long term plan, I hope it gets implemented soon.

In the meantime you don't want to merge this? 🥺

simonsan (approved):

PR #180, opened 2021-01-06
ci: rename job

MarcoIeni:


simonsan (approved):
Can we move all markdown checks into a markdown test and the mdbook stuff into an the same file

PR #183, opened 2021-01-06
add markdown rules

MarcoIeni:
I reviewed all the rules in the url I put in the yaml file.
This is what I decided to add. It is consistent to what we already have by the way.


simonsan:
Could you please comment on each line what each rule does, like in code, so we can actually have an overview for each rule without opening another website to look at?

simonsan (approved):

PR #184, opened 2021-01-06
create separated page in functional programming

### LINKED ISSUE #141: Make the display of a new chapter more clear to the reader
> when walking the chapters using the next button, I initially thought the Functional Programming chapter was part of the anti-patterns chapter and was confused why FP was an anti-pattern.

src.: [reddit](https://www.reddit.com/r/rust/comments/kowtqn/rust_design_patterns_now_also_as_a_book/ghvvj0s/?utm_source=reddit&utm_medium=web2x&context=3)

MarcoIeni:
fix #141


functional/declarative.md
```diff
@@ -0,0 +1,64 @@
+# Declarative vs Imperative
```
simonsan, line 1: Would call it `fp-paradigms.md` and the headline `Function Programming paradigms` as well. In the `SUMMARY` I would call it just `Paradigms` as it's already a subsection of FP.

simonsan: thoughts? @MarcoIeni

MarcoIeni: if you do it like that it looks like "imperative" is a fp paradigm

MarcoIeni: Anyway, I agree that current names are not that good

simonsan: Ah read it yesterday and forgot to answer, sorry. What would be your proposal?

MarcoIeni: The headline and summary title could be `Declarive paradigm`?

pickfire: No, this should be called paradigms. If we put `Declarative paradigm` here and we talk about imperative, it is a bit confusing, so better just put the general world that covers both.

MarcoIeni: Ok, should we go with `Programming paradigms`? We can also link [this](https://en.wikipedia.org/wiki/Programming_paradigm) in the introduction

MarcoIeni: simonsan, tell me if it looks ok now :)

SUMMARY.md
```diff
@@ -40,6 +40,7 @@
   - [Deref Polymorphism](./anti_patterns/deref.md)
 
 - [Functional Programming](./functional/index.md)
+  - [Declarative vs Imperative](./functional/paradigms.md)
```
simonsan, line 43:
```suggestion
  - [Programming paradigms](./functional/paradigms.md)
```

functional/index.md
```diff
@@ -1,66 +1,4 @@
...
-|   3 |   3 |      6 |
-|   6 |   4 |     10 |
-|  10 |   5 |     15 |
-|  15 |   6 |     21 |
-|  21 |   7 |     28 |
-|  28 |   8 |     36 |
-|  36 |   9 |     45 |
-|  45 |  10 |     55 |
+Rust is an imperative language, but it follows many
+[functional programming](https://en.wikipedia.org/wiki/Functional_programming) paradigms.
```
simonsan, line 4: I think this intro should link at least also to the content of the folder itself, like a TOC, in this case just `./paradigms.md`.

Also it should explain what this folder will contain, so i feel this sentence is a bit short? Maybe we write some basic introductory sentences as in the other `./**/intro.md`s for that corresponding subtopics?

functional/declarative.md
simonsan: reviewed it

SUMMARY.md
MarcoIeni: Maybe better "Programming paradigms" as the page title?

functional/index.md
MarcoIeni: > I think this intro should link at least also to the content of the folder itself

I don't like it, the toc is already on the left. The rust book doesn't do that for example.

> Also it should explain what this folder will contain

Yes, a better intro is missing.

simonsan: I thought we did it for the top layer as well, but indeed, we did it only for the second layer (FFI) so yeah `graphs` would be a candidate then for a TOC. my bad :)

SUMMARY.md
simonsan: Yeah, would be clearer, I thought that it looks repeated, because over it will be `Funtional Programming` and then the next line starts with `Programming Paradigms` again. Like Programming ... Programming.

MarcoIeni: done in 0e89dacb7

functional/index.md
MarcoIeni: I added an introduction (by copy pasting from wikipedia ^^'). Tell me if it is ok now.

functional/paradigms.md
```diff
@@ -0,0 +1,66 @@
...
+|   8 |    36 |
+|   9 |    45 |
+|  10 |    55 |
+
+This is how most of us start out programming. We learn that a program is a set of steps.
+
+## Declarative
+
+```rust
+println!("{}", (1..11).fold(0, |a, b| a + b));
```
pickfire, line 41: I think we should mention that one can just `(1..11).sum()`.
```suggestion
println!("{}", (1..11).fold(0, |a, b| a + b)); // `.sum()` is a better alternative, this is just for illustration
```

functional/paradigms.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Whoa! This is really different! What's going on here?
+Remember that with declarative programs we are describing __what__ to do, rather than __how__ to do it.
+`fold` is a function that [composes](https://en.wikipedia.org/wiki/Function_composition) functions.
+The name is a convention from Haskell.
+
+Here, we are composing functions of addition (this closure: `|a, b| a + b)`) with a range from 1 to 10.
```
pickfire, line 49:
```suggestion
Here, we are composing functions of addition (this closure: `|a, b| a + b`) with a range from 1 to 10.
```
Extra closing bracket.

functional/paradigms.md
```diff
@@ -0,0 +1,66 @@
...
+The `0` is the starting point, so `a` is `0` at first.
```
pickfire, line 50: Even though it is being moved. I think it is better to use `acc` and `x` to follow the standard library docs and being easier to read rather than `a` and `b`.

pickfire (approved):
Other than the mentioned comments other parts looks good to me, even though the mentioned comments are unrelated.

MarcoIeni:
Thank you for reviewing this, pickfire.
I opened #207 to track the changes you proposed, I hope you don't mind, but it's better to keep PRs small and on focus.
I like your proposals, we can work on them once simonsan merges this. (I can't merge this because I opened it 😅).

pickfire (approved):
Looks good to me then since it does what this PR mentioned. Heh, @MarcoIeni you are stuck in an issue similar to me, luckily the other reviewer is fast, in the other project the other reviewers are not so fast.

simonsan (approved):

PR #190, opened 2021-01-07
Revise the FFI chapters

### LINKED ISSUE #171: FFI Best Practices Examples Contain Unsoundness and Other Not Great Things
Per a comment by @simonsan, here is an issue to discuss some changes to #106.

**EDIT of maintainer:** Feedback of @Michael-F-Bryan can be found here in the PR https://github.com/rust-unofficial/patterns/pull/106#pullrequestreview-561849641

It seems while I was away, the PR was merged with additional fixes. Only _after_ that happened, did @Michael-F-Bryan come along and do a deeper code review than the initial reviewers did.

It is definitely better late than never. I am happy to make changes, but it will take some time. I do not always get time for this hobby of mine, and the original work was done when I had a lot more than I will for a while.

Anyone who wants to beat me to it can open a PR for anything they think is sufficiently glaring. I would be happy to look at anything, and will probably approve unless it messes up the example.

---

For my own part, here is a brief explanation of where all my work came from.

The lessons I wrote -- and the structure of the code examples -- came from a stand-alone hobby project I did to learn Rust several years ago. It was quite an ambitious one, and I am still surprised how far I got.

I converted most of a 15,000 line abandon-ware C library into Rust, while keeping it _perfectly ABI compatible_ with the original. I ran it through a C-to-Rust translation tool, got it barely working in 100% `unsafe` Rust, and started working my way toward safety.

In so doing, I learned a lot -- but also had some special challenges, and almost certainly missed some things.

I ended up relying on Rust UB in a couple of places specifically because the ABI made me. Even though it is UB, it was often predictable, easy to debug, and paper over. I stopped worrying about things that didn't fall into certain traps in C. In the course of creating these examples, those flaws in the original sneak through sometimes.

The code was also very old, around the Rust 1.15 timeframe. In the process of writing the code examples, I had to do simplifications and modernizations from pre-edition 2018 Rust. It does not surprise me that, when I was unable to test the code I wrote "by eye", it has issues.

---

With that explanation out of the way, I will start with the original comments @Michael-F-Bryan left.

>  * Unwinding across the FFI boundary is UB, so all calls to non-trivial Rust code in FFI functions must be wrapped in std::panic::catch_unwind()

This is true, and a good point. Perhaps it should be an additional best practice?

I also wonder if it wouldn't clutter some of the examples up more, but perhaps that "realism" is worth it.

>  * Almost all of your FFI bindings should unsafe fn's because their memory safety depends on the caller doing the right thing, so not marking them as unsafe is technically unsound

Also a good point. I was less rigorous about this during that project due to the way that project developed, where `unsafe fn` became a marker for "you don't know where all your unsafe blocks are," a habit which transferred over to this text.

>  * No null pointer checks

I am actually somewhat ambivalent about this.

Among C programmers, there is no simple answer to whether you should write "do not pass NULL pointers" into the documented contract of a function, or put explicit checks everywhere. It really depends on how much you trust your users.

Because the _actual behavior_ that results from null pointer access (despite it being UB) is commonly handled cleanly, and because it was less clutter, I chose the former option.

>  * You can tell some of the code hasn't been compiled and run, for example the get_error_from_ffi() function will always panic because you call CString::new() with a buffer that contains a null (the terminator)

It seems that I used an unstable function, which does not simply stop at the first null byte as I would have expected. I agree that is important to fix. 

>  *  Functions not documenting their safety invariants and whether they borrow/own the values passed in from C

This was also caused by an accident of how this document came about. "Object-based APIs" was going to be a big thing that contained many other things as bullet points. Outside of that, it probably would be a good idea to document.

Michael-F-Bryan:
This is a continuation of #106 which addresses some of the items pointed out in the review.

Fixes #171.


Michael-F-Bryan:
@jhwgh1968 how do you think I should approach the `CString` example in the "accepting strings" chapter? 

I feel like the example is too contrived to show how you might use `CString`. It essentially reduces down to

```rust
fn mylib_log(msg: &str) {
  let msg: String = msg.to_owned();
  crate::log(&msg, ...);
}
```

... which isn't really code that people would write.

From my experience you don't really use `CString` when accepting null-terminated strings. Normally you'd keep them around as a `&CStr` or convert to a `String` with something like `CStr::from_ptr(...).to_str().unwrap().to_owned()`. 

The only example I can think of is when you need to store the string somewhere and keep it around after the function ends, but even then I'd use `CStr::from_raw(...).to_owned()` instead of copying to a `Vec` and calling `CString::new(...).unwrap()`.

jhwgh1968:
> how do you think I should approach the `CString` example in the "accepting strings" chapter?
> 
> I feel like the example is too contrived to show how you might use `CString`. 

It is a little contrived, but I could not do a good translation of what I had.

My goal with that alternative was to illustrate a thought process I actually had in the beginning: always copy all pointers to a Rust-owned value ASAP, and then never touch the pointers again, no matter what type they were. That is what I wanted to discourage.

> The only example I can think of is when you need to store the string somewhere and keep it around after the function ends, but even then I'd use `CStr::from_raw(...).to_owned()` instead of copying to a `Vec` and calling `CString::new(...).unwrap()`.

I had code that did a bunch of pointer arithmetic, so I put it in. But that was because I had C code which did that, and it stayed around for quite some time before I completely rewrote it. I agree that most Rust programmers wouldn't think of something that backwards, and it might not be worth putting in.

In fact, your snippet seems like a good thing to replace my alternative with: "if you really want an owned one, use this. Don't do pointer arithmetic or null byte termination yourself, because it's error prone, and this takes care of it."

> There are actually two orthogonal concepts that we'll want to address in the error handling chapter. One is returning the equivalent of Rust's `Result<T, E>`, and the other is the myriad of different conventions that C uses. The current chapter only handles `Result<(), E>` and doesn't talk about how you return both an error code and a value.

I tried to suggest (and wasn't very clear) that the "value" was what was returned by `db_error_description`. I forgot to put in a `db_parse_error_details` function to match it, which would return a `parse_error` struct by value in order to match it.

> For example, how would write FFI bindings for something like this?
> 
> ```rust
> fn parse_elf(filename: *const c_str) -> Result<ElfBinary, ParseError>;
> 
> struct ElfBinary { ... }
> enum ParseError { ... }
> ```

The pattern I wrote would say one of two things: if it's opaque, return it as a pointer. If it's transparent, use an out parameter and an integer code (either non/zero, or an error code number).

```C
/* elfbin.h */
// if transparent
int elf_bin_new(const char* filename, **ElfBinary out_elf);

// if opaque, returns NULL pointer on error
*ElfBinary elf_bin_new(const char *filename);

// either way, you can get more information about the error message with this
//
// pass a pointer to an ElfBinary that was constructed to get the last error from its call.
// pass NULL pointer in order to get the last error from a global call, like "new."
const char *elf_bin_error(*ElfBinary e);
```

Suffice it to say, I'm trying to avoid doing what some libraries do, which is to overwrite `errno`, or interact with any C global state.

idioms/ffi-accepting-strings.md
```diff
@@ -5,15 +5,27 @@
...
+- C strings can contain any arbitrary non-zero byte while Rust strings must be
+  UTF-8
+- C strings are accessed and manipulated using `unsafe` pointer operations
+  while interactions with Rust strings go through safe methods
+
+The Rust standard library comes with C equivalents of Rust's `String` and `&str`
+called `CString` and `&CStr`.
+
+By using `CString` and `&CStr` we avoid a lot of the complexity and `unsafe`
+code involved in converting between C strings and Rust strings.
```
MarcoIeni, line 25:
```suggestion
The Rust standard library comes with C equivalents of Rust's `String` and `&str`
called `CString` and `&CStr`, that allow us to avoid a lot of the complexity and `unsafe`
code involved in converting between C strings and Rust strings.
```

idioms/ffi-accepting-strings.md
```diff
@@ -22,15 +34,23 @@ pub mod unsafe_module {
 
     // other module content
 
+    /// Log a message at the specified level.
+    ///
+    /// # Safety
+    ///
+    /// It is the caller's guarantee to ensure `msg`
```
MarcoIeni, line 41:
```suggestion
    /// It is the caller's guarantee to ensure `msg`:
```

idioms/ffi-accepting-strings.md
```diff
@@ -22,15 +34,23 @@ pub mod unsafe_module {
...
+    ///
+    /// - is not a null pointer
+    /// - points to valid, initialized data
+    /// - points to memory ending in a null byte
+    /// - won't be mutated for the duration of this function call
     #[no_mangle]
-    pub extern "C" fn mylib_log(msg: *const libc::c_char, level: libc::c_int) {
+    pub unsafe extern "C" fn mylib_log(msg: *const libc::c_char, level: libc::c_int) {
```
MarcoIeni, line 48: I don't like that we are extending the scope of unsafe.
`crate::log` is a safe function, so there is no point in placing it inside unsafe, right?
Maybe we could extract `match` into its own unsafe function?

MarcoIeni:
> always copy all pointers to a Rust-owned value ASAP, and then never touch the pointers again, no matter what type they were. That is what I wanted to discourage.

This is quite nice, we want to explain it in the book?

Anyway here you are talking about another part of the book (error handling), which is not really related to this PR, right?
Because if you want once we solve the comments I left we could merge this and address other problems in separate PRs/issues.

idioms/ffi-accepting-strings.md
jhwgh1968: The reason is because [`#[no_mangle]` is inherently unsafe, due to implied guarantees about the rest of the code](https://github.com/rust-lang/rust/issues/28179). It is those guarantees that the `unsafe fn` represents.

jhwgh1968:
> > always copy all pointers to a Rust-owned value ASAP, and then never touch the pointers again, no matter what type they were. That is what I wanted to discourage.

> Anyway here you are talking about another part of the book (error handling), which is not really related to this PR, right?

I was talking about what motivated a previous example in Accepting Strings, which in retrospect was very convoluted.

It is a general idea, but Accepting Strings was where I put it, because that's where it was most notable in my early code. Strings come in as `const c_char *`, and I was copying them into Rust owned `String`s. I knew it was a lot of work, but I thought it was "safer," when I was worried about splitting ownership of the old C code. That is what I wanted to discourage.

I do not know whether it is worth a mention in Accepting Strings or not, if it is moved to a general principle. I would approve this PR either way.

idioms/ffi-accepting-strings.md
MarcoIeni: Thank you, I learned something :)
Marking this as solved.

simonsan:
@jhwgh1968 @Michael-F-Bryan Hey you both, any updates on this? Asking because we're probably changing the `patterns` subfolder structure so `ffi` has it's own subfolder, this would make a rebase needed probably with some conflicts to solve. Just for letting you know. ;)

Michael-F-Bryan:
I'm going to close this PR because I don't have the time to complete it and didn't really want to invest much more than that initial review.

MarcoIeni:
Hi Michael, I reopen this and I assign it to me. I will continue the work you have started :)

idioms/ffi-accepting-strings.md
```diff
@@ -6,17 +6,25 @@ When accepting strings via FFI through pointers, there are two principles that
...
-`to_string()` on the string slice.
+- C strings are null-terminated while Rust strings store their length
+- C strings can contain any arbitrary non-zero byte while Rust strings must be
+  UTF-8
+- C strings are accessed and manipulated using `unsafe` pointer operations
+  while interactions with Rust strings go through safe methods
+
+The Rust standard library comes with C equivalents of Rust's `String` and `&str`
+called `CString` and `&CStr`, that allow us to avoid a lot of the complexity and `unsafe`
+code involved in converting between C strings and Rust strings.
```
MarcoIeni, line 24:
```suggestion
called `CString` and `&CStr`, that allow us to avoid a lot of the complexity
and `unsafe` code involved in converting between C strings and Rust strings.
```

idioms/ffi-accepting-strings.md
```diff
@@ -6,17 +6,25 @@ When accepting strings via FFI through pointers, there are two principles that
 should be followed:
 
 1. Keep foreign strings "borrowed", rather than copying them directly.
-2. Minimize `unsafe` code during the conversion.
+2. Minimize the amount of complexity and `unsafe` code involved in converting
+   from a C-style string to native Rust strings
```
MarcoIeni, line 10:
```suggestion
   from a C-style string to native Rust strings.
```

MarcoIeni (approved):

MarcoIeni:
Merged this! If you think something valuable from the above comments needs to be discussed further, please open an issue or another PR!

PR #200, opened 2021-01-13
Talk about mem::take instead of mem::replace

### LINKED ISSUE #96: mem-take instead of mem-replace
In https://github.com/rust-unofficial/patterns/blob/964dd9932b7bf41b0905c264ea065781efcdb6b7/idioms/mem-replace.md,

`mem::replace(name, String::new())`

can be made more terse by using

`mem::take(name)`

instead. Rust playground link to show it still compiles: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6f29fd3165760bf068a7c4200fa8cbc5

ventlark:
The motive behind this PR is that mem::take is often neater and more elegant than mem::replace, especially in the example listed in the original mem::replace article.

This PR changes the mem::replace article to instead focus on mem::take. The article does mention mem::replace as an alternative if you don't want the default value to be the `src`, or if your type doesn't implement `Default`.

This also fixes a typo in the source code block that prevented the block from being compiled and executed, and then fixes a compile error in the block itself.

EDIT of Maintainer: Closes #96

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -8,7 +8,7 @@
   - [The `Default` Trait](./idioms/default.md)
   - [Collections Are Smart Pointers](./idioms/deref.md)
   - [Finalisation in Destructors](./idioms/dtor-finally.md)
-  - [`mem::replace(_)`](./idioms/mem-replace.md)
+  - [`mem::{take(_), replace(_)}`](./idioms/mem-replace.md)
   - [On-Stack Dynamic Dispatch](./idioms/on-stack-dyn-dispatch.md)
   - [Foreign function interface usage](./idioms/ffi-intro.md)
     - [Idiomatic Errors](./idioms/ffi-errors.md)
--- a/idioms/mem-replace.md
+++ b/idioms/mem-replace.md
@@ -38,7 +38,7 @@ fn a_to_b(e: &mut MyEnum) {
 
 This also works with more variants:
 
-```Rust
+```rust
 use std::mem;
 
 enum MultiVariateEnum {
@@ -49,7 +49,7 @@ enum MultiVariateEnum {
 }
 
 fn swizzle(e: &mut MultiVariateEnum) {
-    use self::MultiVariateEnum::*;
+    use MultiVariateEnum::*;
     *e = match *e {
         // Ownership rules do not allow taking `name` by value, but we cannot
         // take the value out of a mutable reference, unless we replace it:
--- a/idioms/mem-replace.md
+++ b/idioms/mem-replace.md
@@ -1,4 +1,4 @@
-# `mem::replace` to keep owned values in changed enums
+# `mem::take` to keep owned values in changed enums
 
 ## Description
 
@@ -29,7 +29,7 @@ fn a_to_b(e: &mut MyEnum) {
         // (note that empty strings don't allocate).
         // Then, construct the new enum variant (which will
         // be assigned to `*e`, because it is the result of the `if let` expression).
-        MyEnum::B { name: mem::replace(name, String::new()) }
+        MyEnum::B { name: mem::take(name) }
 
     // In all other cases, we return immediately, thus skipping the assignment
     } else { return }
@@ -53,8 +53,8 @@ fn swizzle(e: &mut MultiVariateEnum) {
     *e = match *e {
         // Ownership rules do not allow taking `name` by value, but we cannot
         // take the value out of a mutable reference, unless we replace it:
-        A { ref mut name } => B { name: mem::replace(name, String::new()) },
-        B { ref mut name } => A { name: mem::replace(name, String::new()) },
+        A { ref mut name } => B { name: mem::take(name) },
+        B { ref mut name } => A { name: mem::take(name) },
         C => D,
         D => C
     }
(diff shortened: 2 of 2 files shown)


ventlark:
#96 can be closed with this

idioms/mem-take.md
```diff
@@ -95,6 +99,10 @@ borrow checker. The compiler may fail to optimize away the double store,
 resulting in reduced performance as opposed to what you'd do in unsafe
 languages.
 
+The type you are taking needs to implement the `Default` trait.
```
simonsan, line 102:
```suggestion
The type you are taking needs to implement the [`Default` trait](./default.md).
```

SUMMARY.md
```diff
@@ -8,7 +8,7 @@
   - [The `Default` Trait](./idioms/default.md)
   - [Collections Are Smart Pointers](./idioms/deref.md)
   - [Finalisation in Destructors](./idioms/dtor-finally.md)
-  - [`mem::replace(_)`](./idioms/mem-replace.md)
+  - [`mem::take(_)`](./idioms/mem-take.md)
```
simonsan, line 11: I'm not so sure here if this is the right way. This is not really backwards compatible, so for example people bookmarking this idiom for later, will get a 404 for this now. Not sure if it would be better to keep the name for the markdown file would be better (so links stay the same), but changing the title to `mem::{replace(_), take(_)}` would be more sufficient. Let's discuss this here.

idioms/mem-take.md
```diff
@@ -75,10 +75,14 @@ into our `MyEnum::B`, but that would be an instance of the [Clone to satisfy
...
-`mem::replace` lets us swap out the value, replacing it with something else. In
-this case, we put in an empty `String`, which does not need to allocate. As a
-result, we get the original `name` *as an owned value*. We can then wrap this in
-another enum.
+`mem::take` lets us swap out the value, replacing it with it's default value,
+and returning the previous value. For `String`, the default value is an empty
+`String`, which does not need to allocate. As a result, we get the original
+`name` *as an owned value*. We can then wrap this in another enum.
+
+`mem::replace` is very similar, but allows us to specify what to
```
simonsan, line 83:
```suggestion
__NOTE:__ `mem::replace` is very similar, but allows us to specify what to
```

idioms/mem-take.md
```diff
@@ -95,6 +99,10 @@ borrow checker. The compiler may fail to optimize away the double store,
 resulting in reduced performance as opposed to what you'd do in unsafe
 languages.
 
+The type you are taking needs to implement the [`Default` trait](./default.md).
```
simonsan, line 102:
```suggestion
Furthermore the type you are taking needs to implement the [`Default` trait](./default.md).
```

simonsan (changes requested):

idioms/mem-take.md
ventlark: Committed in 0eaa3ba, with an added comma.

SUMMARY.md
ventlark: I agree. Rebased so the filename change never happened, and changed title to `mem::{take(_), replace(_)}`.

simonsan (approved):

ventlark:
Sorry for the ping to re-review. I though that was approved before my changes.

simonsan (approved):

simonsan:
> Sorry, ignore the ping to re-review. I though that was approved before my changes.

No problem. ;-)

MarcoIeni:
Should we rename the file to mem-take.md?
Other than this it looks good to me :)

simonsan:
> Should we rename the file to mem-take.md?
> Other than this it looks good to me :)

https://github.com/rust-unofficial/patterns/pull/200#discussion_r556477705

Talked about it here

fade2black:
Regarding `MyEnum ` example. Comments says that 

> this takes out our `name` and put in an empty String instead
> (note that empty strings don't allocate).
>Then, construct the new enum variant (which will
> be assigned to `*e`, because it is the result of the `if let` expression).

I find _(note that empty strings don't allocate)_ part a bit misleading. As if, if empty string would allocate then we couldn't write like this. Wouldn't it dropped even if it allocated? Just after `*e = if let...` assignment. 
I replaced `MyEnum::B { name: mem::take(name) }` with `MyEnum::B { name: mem::replace(name, "some-str".to_string()) }`. It works fine. But I agree writing `MyEnum::B { name: mem::replace(name, "some-str".to_string()) }` is a nonsense.

pickfire:
> I find (note that empty strings don't allocate) part a bit misleading. As if, if empty string allocated then we couldn't write like this. Wouldn't the allocated mem be dropped even if it allocated? Just after *e = if let... assignment.

From the docs.

>Creates a new empty String.
>
>Given that the String is empty, this will not allocate any initial buffer. While that means that this initial operation is very inexpensive, it may cause excessive allocation later when you add data. If you have an idea of how much data the String will hold, consider the with_capacity method to prevent excessive re-allocation.

Empty String does not allocate.

I don't understand why is that a nonsense.

fade2black:
@pickfire 
When I read `(note that empty strings don't allocate)` it seemed to me a sort of warning, and first thing came to my mind "why note? won't it work if it allocates?".  I tried `replace` instead of `take`. It worked.  

>I don't understand why is that a nonsense.
I mean writing `MyEnum::B { name: mem::replace(name, "some-str".to_string()) }` instead of `MyEnum::B { name: mem::take(name) }` is meaningless. The right thing is to take and leave an empty string which does not allocate. Why would we replace with another string if it is going to be dropped after all?

PR #201, opened 2021-01-13
CONTRIBUTING: add style guide

### LINKED ISSUE #196: Formalize language style
With more contributors, more differences in writing style will become apparent. This issue is a proposal to start a guideline to be able to refer back to for consistent writing.

Most words if not all, seem to use the British style (e.g. -ise endings such as 'finalise'; also used in Australia and New Zealand), to name just one example. One may out of habit (or stubbornness) use US-style. Using US-style could be discussed, but it's important to have agreement upon consistency.

A good example of inconsistent language usage is the usage of "we" and "you":

> **We can avoid** this by using &str instead, and letting &String coerce to a &str whenever the function is invoked.
>
>Source: _idioms/coercion-arguments_

> This pattern is only of interest in Rust. In GC'd languages, **you'd take** the reference to the value by default (and the GC would keep track of refs), and in other low-level languages like C **you'd simply alias** the pointer and fix things later.
>
>However, in Rust, **we have to** do a little more work to do this. An owned value may only have one owner, so to take it out, **we need to** put something back in
>
>Source: _idioms/mem-replace_

One could argue that this is done on purpose in this specific instance, but it felt out of place when I read it. But there's also the usage of "me":

>The etymology here is **unclear to me**. The terms 'fold' and 'folder' are used in the Rust compiler, although **it appears to me** to be more like a map than a fold in the usual sense. See the discussion below for more details.
>
>Source: _patterns/fold_

Most of the text is also written informally (e.g. 'anyways' over 'anyway') and with abbreviations (in this instance of 'GC', even without explanation).

I am not necessarily suggesting a change and lots of things could be considered gray area (grey area?), but it would be good if there is some writing guideline to be able to refer to when in doubt.

MarcoIeni:
Closes #196

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ Please make `Draft Pull requests` early so we can follow your progress and can g
 - Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
 - Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). Tl;dr:
   - Prefer full types name. For example `Option<T>` instead of `Option`.
-  - Prefer line comments (`//`) over block comments (`/* */`).
+  - Prefer line comments (`//`) over block comments (`/* */`) where applicable.
 
 ## Check the article locally
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,8 +47,8 @@ Please make `Draft Pull requests` early so we can follow your progress and can g
 
 - Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
 - Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). Tl;dr:
-  - Use full types name. For example `Option<T>` instead of `Option`
-  - Prefer line comments (`//`) over block comments (`/* */`)
+  - Prefer full types name. For example `Option<T>` instead of `Option`.
+  - Prefer line comments (`//`) over block comments (`/* */`).
 
 ## Check the article locally
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,8 @@ Please make `Draft Pull requests` early so we can follow your progress and can g
 
 ## Style guide
 
+In order to have a consistent style across the book, we suggest to:
+
 - Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
 - Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). Tl;dr:
   - Prefer full types name. For example `Option<T>` instead of `Option`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
 
 Please make `Draft Pull requests` early so we can follow your progress and can give early feedback (see the following section).
 
+## Style guide
+
+- Follow the book [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
+- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). TL;DR:
+  - Use American English
+  - Use full types name. For example `Option<T>` instead of `Option`
+  - Prefer line comments (`//`) over block comments (`/* */`)
+
 ## Check the article locally
 
 Before submitting the PR launch the commands `mdbook build` to make sure that the book builds and `mdbook test` to make sure that
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ Please make `Draft Pull requests` early so we can follow your progress and can g
 
 ## Style guide
 
-- Follow the book [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
+- Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
 - Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). TL;DR:
   - Use American English
   - Use full types name. For example `Option<T>` instead of `Option`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
 
 Please make `Draft Pull requests` early so we can follow your progress and can give early feedback (see the following section).
 
+## Style guide
+
+- Follow the book [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
+- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). TL;DR:
+  - Use American English
```
pickfire, line 50:
```suggestion
  - Use formal American English
```

CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
...
+  - Use full types name. For example `Option<T>` instead of `Option`
```
pickfire, line 51: I don't think this is necessary.

CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
...
+  - Prefer line comments (`//`) over block comments (`/* */`)
```
pickfire, line 52: I think this is pretty much standard that we don't need to mention this already, I wonder if we have even came across this?

CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
 
 Please make `Draft Pull requests` early so we can follow your progress and can give early feedback (see the following section).
 
+## Style guide
+
+- Follow the book [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
+- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). TL;DR:
```
pickfire, line 49:
```suggestion
- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text).
- tl;dr
```
Would this be better? I don't see that is looks better but both seemed weird.

CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
 
 Please make `Draft Pull requests` early so we can follow your progress and can give early feedback (see the following section).
 
+## Style guide
+
```
simonsan, line 47: Would propose a small introduction text here why we apply a style guide to the docs

CONTRIBUTING.md
```diff
@@ -43,6 +43,14 @@ Don't forget to add your new article to the `SUMMARY.md` to let it be rendered t
...
+- Follow the book [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
```
simonsan, line 48:
```suggestion
- Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
```

CONTRIBUTING.md
simonsan: That is near to impossible for us to check as we are international and not English natives. If so I would prefer British English over American, I don't think this is a good rule, to be honest. Let's discuss about it.

Also we want to make it as easy as possible for people to write and not put more and more rules and everything in the way. So I'm really unsure the direction we are taking with this.

pedzed: It doesn't have to be as formal as you would write to Queen Elizabeth, but it should in my opinion also not be slang or too informal.

simonsan: https://rust-unofficial.github.io/patterns/idioms/mem-replace.html#advantages

> Look ma, no allocation! Also you may feel like Indiana Jones while doing it.

I love that actually and I find it kind of what a collaborative book is. Different styles of different authors and not something that is made out of stone and would always sound the same. For sure if it gets too colloquial/informal we can interfere in the PR process, but imho no need to write that down as a rule. I'm trusting the people maintaining this now and in the long-term future that there will be a rational about how a PR/article is written and if it fits to the books style.

MarcoIeni: I also like informal style, see my comment in the relative issue. I will resolve this conversation without committing this suggestion.

CONTRIBUTING.md
MarcoIeni: Link them to the official documentation? Yes, it would be really cool. If that's what you mean you can suggest a change about this.

CONTRIBUTING.md
MarcoIeni: Yes, it's standard, but it doesn't hurt to put it there if there are new people coming from different languages I think.
Does anybody else have an opinion on this?

CONTRIBUTING.md
MarcoIeni: It looks good to me in the same line as it is now. But maybe it is better to write `Tl;dr` instead of `TL;DR`, right?

CONTRIBUTING.md
MarcoIeni: YES! I talked about this in the issue

CONTRIBUTING.md
MarcoIeni: Actually. now that I think about it we are writing types with the markdown syntax "`".
It would be silly to link to the official documentation every time.

CONTRIBUTING.md
simonsan: i saw block comments used in arguments to keep the argument brackets of functions on the same line. they are sometimes more useful than line comments. I'm not sure if we can adhere to this style completely. But in general it might be not bad to state it maybe just telling the people "hey, look at the markdown of another document and copy the style of the code" or so?

MarcoIeni: > i saw block comments used in arguments to keep the argument brackets of functions on the same line. they are sometimes more useful than line comments. I'm not sure if we can adhere to this style completely.

It is just the preferred style, there can be exceptions.

> hey, look at the markdown of another document and copy the style of the code

I think that's what people do by default.

The reason why I put the TL;DR there is that I wanted to summarize what the RFC says in order to avoid forcing people to read all the RFC and only put what's relevant for us.

Do you agree to keep this line there?

simonsan:
```suggestion
  - Prefer line comments (`//`) over block comments (`/* */`) where applicable
```

Like this maybe?

MarcoIeni: fixed in https://github.com/rust-unofficial/patterns/pull/201/commits/50c066357eab9601ae297b5adb62444d2fd5742c

CONTRIBUTING.md
MarcoIeni: done in https://github.com/rust-unofficial/patterns/pull/201/commits/777a0f73e56263987f3392536be217d934ac7ed7
feel free to suggest changes

simonsan (approved):

PR #208, opened 2021-01-21
Shorten lines to line-length 80

### LINKED ISSUE #185: Cut long lines
In [this](https://github.com/rust-unofficial/patterns/blob/master/.markdownlint.yaml) file we set the maximum line length for our markdown linter.
Our end goal is to bring it to something sane, like 150 for example.
Right now it is 300.

You should edit that file in order to decrease the number a little bit and fix all the errors that are showed by running `markdownlint` as explained in [CONTRIBUTING](https://github.com/rust-unofficial/patterns/blob/master/CONTRIBUTING.md#markdown-lint).

Try to keep PR small, i.e. do not set that number to 150 directly and fix hundreds of lines at the same time.
Fixing around 10 lines per PR might be a good number for example, so you can set it to 230 for the first iteration for example.

You can see [this](https://github.com/rust-unofficial/patterns/pull/176) PR as an example.
If you want to work on this leave a comment and we will assign the issue to you.
You can also do just one single PR, that's fine, you don't have to do all the job. Just specify that in the comment :)

simonsan:
Closes #185


idioms/priv-extend.md
```diff
@@ -28,15 +28,19 @@ fn main(s: a::S) {
...
+ensuring that the struct is future-proof.
+
+The downside of this approach is that you might need to add an otherwise unneeded
+field to the struct. You can use the `()` type so that there is no runtime overhead
+and prepend `_` to the field name to avoid the unused field warning.
+
+If Rust allowed private variants of enums, we could use the same trick to make
+adding a variant to an enum backwards compatible. The problem there is exhaustive
+match expressions. A private variant would force clients to have a `_` wildcard
+pattern. A common way to implement this instead is using the [`#[non_exhaustive`]](<https://doc.rust-lang.org/reference/attributes/type_system.html>)
```
simonsan, line 45:
```suggestion
pattern. A common way to implement this instead is using the [`#[non_exhaustive`]](https://doc.rust-lang.org/reference/attributes/type_system.html)
```

simonsan:
```suggestion
pattern. A common way to implement this instead is using the [`#[non_exhaustive`]](<https://doc.rust-lang.org/reference/attributes/type_system.html>)
```

simonsan:
```suggestion
pattern. A common way to implement this instead is using the [`#[non_exhaustive`]](<https://doc.rust-lang.org/reference/attributes/type_system.html>)
```

CONTRIBUTING.md
```diff
@@ -2,66 +2,81 @@
...
+Please make `Draft Pull requests` early so we can follow your progress and can
+give early feedback (see the following section).
 
 ## Style guide
 
 In order to have a consistent style across the book, we suggest to:
 
 - Follow the official Rust book's [style guide](https://github.com/rust-lang/book/blob/master/style-guide.md).
-- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text). Tl;dr:
+- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text)
```
MarcoIeni, line 64:
```suggestion
- Follow [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text).
```

idioms/priv-extend.md
MarcoIeni, line 45:
```suggestion
pattern. A common way to implement this instead is using the [`#[non_exhaustive]`](<https://doc.rust-lang.org/reference/attributes/type_system.html>)
```

MarcoIeni: Why the angle brackets?

patterns/compose-structs.md
```diff
@@ -33,7 +33,8 @@ fn baz(a: &mut A) {
     // The later usage of x causes a to be borrowed for the rest of the function.
     let x = foo(a);
     // Borrow checker error:
-    // let y = bar(a); // ~ ERROR: cannot borrow `*a` as mutable more than once at a time
+    // let y = bar(a); // ~ ERROR: cannot borrow `*a` as mutable more than once
+    // at a time
```
MarcoIeni, line 37:
```suggestion
                       //          at a time
```

MarcoIeni: better alignment. If you don't like this feel free to refuse this.

MarcoIeni:
Apart for these comments it looks good to me.
I hope I haven't missed anything.

idioms/priv-extend.md
simonsan: because ci doesn't allow bare urls in this case, don't ask me why

MarcoIeni: ok, got it. Anyway, take a look at the suggestion, too.

simonsan:
@MarcoIeni Before merging, can you please remove all the text in the message box down under the merge commit's title, it's not useful in most cases, and for sure also not here with all the commits.

MarcoIeni (approved):

MarcoIeni:
> all the text in the message box down under the merge commit's title

What do you mean? Can you take a screenshot?

PR #216, opened 2021-02-02
Add link to a list of all clippy lints

simonsan:


pickfire:
How is it related to the context?

simonsan:
> How is it related to the context?

It's a list of lints with filters, so you can easily filter out deprecated ones or for groups. Good to have an overview.

MarcoIeni (approved):

PR #245, opened 2021-03-08
Fix format by moving advantage to new bullet

dariuswiles:
It looks like the third advantage was accidentally added to the end of the second, so I added a return character to add it to a new bullet point.


simonsan (approved):
Thank you! :)

PR #250, opened 2021-03-29
[WIP] add fallible closure

jRimbault:
This is a short idioms I think I've seen several times.

Using a fallible closure inside an infallible function to allow the use of the `?` operator.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -19,6 +19,7 @@
   - [Privacy For Extensibility](./idioms/priv-extend.md)
   - [Easy doc initialization](./idioms/rustdoc-init.md)
   - [Temporary mutability](./idioms/temporary-mutability.md)
+  - [Fallible closure](./idioms/fallible-closure.md)
 
 - [Design Patterns](./patterns/index.md)
   - [Behavioural](./patterns/behavioural/intro.md)
--- a/idioms/fallible-closure.md
+++ b/idioms/fallible-closure.md
@@ -1,0 +1,36 @@
+# Fallible closure inside infallible function
+
+## Description
+
+Build a fallible closure inside of an apparently infallible function.
+
+## Example
+
+```rust
+use std::path::Path;
+
+// this function hides its failure states
+fn file_stem(path: &Path) -> &str {
+    // this closure is fallible, so it can make use of `?`
+    // fn() -> Option<&str>
+    let inner = || Some(path.file_stem()?.to_str()?);
+    inner().unwrap_or("untitled")
+}
+```
+
+## Motivation
+
+You may sometimes want to use the [`?`] operator inside an infallible function,
+in those case a simple closure inside your function could be the simple solution.
+
+## Advantages
+
+This allows using the terser syntax permitted by the [`?`] operator inside an
+infallible context.
+
+## Disadvantages
+
+This is hiding failure states from the consumer of the API.
+
+
+[`?`]: https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator


idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+    // fn() -> Option<&str>
+    let inner = || Some(path.file_stem()?.to_str()?);
+    inner().unwrap_or("untitled")
+}
+```
+
+## Motivation
+
+You may sometimes want to use the [`?`] operator inside an infallible function,
+in those case a simple closure inside your function could be the simple solution.
```
simonsan, line 24:
```suggestion
in those case a closure inside your function could be the simple solution.
```

simonsan:
@jRimbault Do you still want to add anything or is it ready to review?

jRimbault:
I think it would be good to elaborate a bit more and give a better example, but yes I'm ready to review and change things if needed ?

jRimbault:
This was prompted by someone on reddit asking how to write [the first two functions][playground], to which I replied with the 3rd implementation. I think I've seen that idiom here and there in Rust code.


[playground]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=28ccd52e8f4d87d3ee9ff3547339dcbf

SUMMARY.md
```diff
@@ -19,6 +19,7 @@
   - [Privacy For Extensibility](./idioms/priv-extend.md)
   - [Easy doc initialization](./idioms/rustdoc-init.md)
   - [Temporary mutability](./idioms/temporary-mutability.md)
+  - [Fallible closure](./idioms/fallible-closure.md)
```
simonsan, line 22: We might find a better headline for the summary to let this idiom be found more easily.

idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+    // fn() -> Option<&str>
+    let inner = || Some(path.file_stem()?.to_str()?);
+    inner().unwrap_or("untitled")
+}
+```
+
+## Motivation
+
+You may sometimes want to use the [`?`] operator inside an infallible function,
+in those case a closure inside your function could be the simple solution.
```
simonsan, line 24:
```suggestion
in those cases a closure inside your function could be the simple solution.
```

SUMMARY.md
jRimbault: I agree, this often comes up when people want to avoid "arrow code" with let guards or "match hell", maybe something around that ?

simonsan:
The general question is also when you should do it and when not. I mean it's obviously pretty bad for error handling and I would consider it kind of an anti-pattern somehow in this regards, as also mentioned in the `# Disadvantages`. Also it's quite unclear (for me as a reader) why you would use `try/?` to bubble up errors, but then don't handle them. Maybe to just sketch out/hack something fastly? I think in this regards the text could elaborate a bit more, imho.

idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Motivation
+
+You may sometimes want to use the [`?`] operator inside an infallible function,
+in those cases a closure inside your function could be the simple solution.
+
+## Advantages
+
+This allows using the terser syntax permitted by the [`?`] operator inside an
+infallible context.
```
simonsan, line 29: Bit repetitive, `it's good for this, because it's good` like. maybe a bit more elaboration. maybe the text from my last comments would fit better in here then

idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Disadvantages
+
+This is hiding failure states from the consumer of the API.
```
simonsan, line 33: why is that bad and when do you really want to avoid it? mentioning some cases would be good and probably also state, that if this is overused it could be considered an `anti-pattern`

jRimbault:
I'll sleep on on your advice (it's 11pm in my timezone). I think I should first find some good examples (of good* and bad use cases). I've only had [one] legit use case. It's often came up when trying to help someone else though. IMO this should definitely not be in a function in a public API. It's really about syntax convenience when the author thinks they have a good default case.


[one]: https://github.com/jRimbault/yadf/blob/87004edf35bfa1757b52062e7eaebd5b23ffa370/src/fs/mod.rs#L37-L65

idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+    // fn() -> Option<&str>
+    let inner = || Some(path.file_stem()?.to_str()?);
+    inner().unwrap_or("untitled")
+}
+```
+
+## Motivation
+
+You may sometimes want to use the [`?`] operator inside an infallible function,
+in those cases a closure inside your function could be the simple solution.
```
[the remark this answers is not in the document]
pickfire: Is this even an idiom? I thought it is just a syntax?

idioms/fallible-closure.md
```diff
@@ -0,0 +1,35 @@
...
+## Example
+
+```rust
+use std::path::Path;
+
+// this function hides its failure states
+fn file_stem(path: &Path) -> &str {
+    // this closure is fallible, so it can make use of `?`
+    // fn() -> Option<&str>
+    let inner = || Some(path.file_stem()?.to_str()?);
```
pickfire, line 16: I believe this example is not making full use of `?`, it could even have `TryInto` IIRC.

jRimbault: It could be written :

```rust
let inner =  || path.file_stem()?.to_str();
```

but that wouldn't show much

pickfire: I do think that is better. I don't know why we need to have another layer for that.

PR #262, opened 2021-07-08
Note on Vec

### LINKED ISSUE #261: In `Collections are Smart Pointers` data is `T` and not `[T]`
Hey! Lovely book but in [Collections are Smart Pointers](https://rust-unofficial.github.io/patterns/idioms/deref.html) there
is
```rust

struct Vec<T> {
    data: T,
    //..
}
```

But shouldn't it be like 

```rust 
struct Vec<T> {
  data: [T], 
  // or
  data: RawVec<T>,
```

Or something like that?

pickfire:
Fix #261

cc @CGMossa

### DIFF
--- a/idioms/deref.md
+++ b/idioms/deref.md
@@ -11,7 +11,7 @@ and borrowed views of data.
 use std::ops::Deref;
 
 struct Vec<T> {
-    data: T,
+    data: RawVec<T>,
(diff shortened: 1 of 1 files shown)


MarcoIeni (approved):
Thanks!

PR #273, opened 2021-09-15
Escape underscores in a link in functional/generics-type-classes.md

WaffleLapkin:
Previously they were parsed as italic modifiers.


simonsan (approved):

simonsan:
Thank you!

PR #282, opened 2021-10-17
deny-warnings: remove contradiction with rustc doc

### LINKED ISSUE #281: Contradiction with the rustc book on --cap-lints and forbid
https://github.com/rust-unofficial/patterns/blob/66d7e6c7daea2b6cf50f6cb230c7203ece0d8678/anti_patterns/deny-warnings.md#L39-L42

This seems to contradict this:

https://github.com/rust-lang/rust/blob/3bf55753c680e653eaa9d11dd66e56844bd04d32/src/doc/rustc/src/lints/levels.md#L102-L105

> However, lint levels may still be capped with `--cap-lints` (see below) so `rustc --cap-lints warn` will make lints set to 'forbid' just warn.

marcoieni:
Closes #281

As the issue says, the `forbid` acts the same as `deny` with the --cap-lints` flag, so there is no need to specify this special case.
I just removed the sentence for the sake of simplicity.


simonsan:
Thank you :)

Issue #145, opened 2021-01-03
Add "Use Self in impl" idiom


MarcoIeni:
Idea from #61
Should we add an idiom that suggest to use `Self` inside `impl`? See the PR above for an example.

- Disadvantages: readability
- Advantages: refactoring

pickfire:
There is a reason this is not added last time and left hanging, it is due to no consensus on whether `Self` is better, personally I would only try to use `Self` when the type is very complicated (so in this case using `Self` is easier to read) but I use the original type in almost any other cases to make it easier to read.

I don't think it is a good idea to write it in the idiom part since it makes not doing so looks bad.

simonsan:
I think even stating what you just said, comparing it and giving information on that matter is already valuable enough to have it in the repository. To give people the knowledge of other people at hand they can think about adopting. These are no rules.

I see this repository more as a `collection of collaborative collective knowledge` than "a statement of what people need to do and what not". So imho it's not really about if `Self` or `FullType` is better but __in which situations is each of it good to use and where not__.

MarcoIeni:
Ok, so basically our decision is to write this including also the content of the comment of pickfire, right?
If you want to write it leave a comment!

pickfire:
I wonder if we should write this? Or maybe we can just keep note of this while writing other examples?

MarcoIeni:
> I wonder if we should write this? Or maybe we can just keep note of this while writing other examples?

IMHO other examples should link to this idiom. Otherwise you duplicate the explanation of this concept all over the place.

joshka:
There's a clippy lint for this (in nursery) https://rust-lang.github.io/rust-clippy/master/index.html#use_self

I'd be inclined to move all code to rust projects (https://github.com/rust-unofficial/patterns/issues/118) and then turn on lints including this one and have all the code use Self as a standard. I wouldn't generally duplicate clippy lints unless there's extra info to add.

PR #284, opened 2021-12-14
chore: use `Self` instead of the struct name

sassman:
This PR just fixes a minor but IMO important detail: whenever returning with `new` or `default` using `Self` is preferable

### DIFF
--- a/patterns/behavioural/newtype.md
+++ b/patterns/behavioural/newtype.md
@@ -35,7 +35,7 @@ impl Bar {
     // Constructor.
     pub fn new(
         //..
-    ) -> Bar {
+    ) -> Self {
 
         //..
 


wookietreiber:
**clippy** has a check for this, although in nursery ATM, see https://rust-lang.github.io/rust-clippy/master/index.html#use_self

Personally, I'm all  for using `Self`, so :+1:

wookietreiber (approved):

MarcoIeni (approved):
I agree, thanks!

pickfire:
I don't think this is a good idea given that it reduces the clarity, I think that's the reason why it is still in nursery.

sassman:
> I don't think this is a good idea given that it reduces the clarity, I think that's the reason why it is still in nursery.

Could you please elaborate what exactly would reduce the clarity? 

I'm having a hard time to imagine what clarity you might refer to.

pickfire:
See https://github.com/rust-unofficial/patterns/issues/145

pickfire:
Can we revert this pull request given that it is still in nursery and no benefits that using `Self` is better?

MarcoIeni:
Hey pickfire, I think that using `Self` here is fine, since it's just an example code.
That issue is more for the page dedicated to that idiom, I think.
Also there is the benefit that when you copy-paste that piece of code you don't have to change the struct name, because `Self` will work for you.
I hope you don't mind that much :)

pickfire:
>  Also there is the benefit that when you copy-paste that piece of code you don't have to change the struct name, because Self will work for you.

If your file only have a single `impl` and is short then yeah, no issue. Imagine if your files is 1000 lines and tons of `impl` for a single enum or struct, then you can easily get lost on the context, having the return there can give more context.

PR #298, opened 2022-02-27
change to introductory text on design patterns

MarkoM-dot:
A change to the intro text on Design Patterns. Open to suggestions!

### DIFF
--- a/intro.md
+++ b/intro.md
@@ -7,10 +7,14 @@ If you are interested in contributing to this book, check out the
 
 ## Design patterns
 
-When developing programs, we have to solve many problems.
-A program can be viewed as a solution to a problem.
-It can also be viewed as a collection of solutions to many problems.
-All of these solutions work together to solve a bigger problem.
+In software development, we often come across problems that share similarities
+regardless of the environment they appear in. Although the implementation details
+are crucial to solve the task at hand, we may abstract from these particularities to find
+some nugget of valuable practices that have a more general application.
+Design patterns are a collection of reusable tried and tested solutions to recurring
+problems in engineering. They make our software more modular, maintainable, and extensible.
+Moreover, these patterns provide a common language for developers to use, making them
+an excellent tool for effective communication when problem-solving in teams.
 
 ## Design patterns in Rust
 


intro.md
```diff
@@ -7,10 +7,14 @@ If you are interested in contributing to this book, check out the
...
-It can also be viewed as a collection of solutions to many problems.
-All of these solutions work together to solve a bigger problem.
+In software development, we often come across problems that share similarities
+regardless of the environment they appear in. Although the implementation details
+are crucial to solve the task at hand, we may abstract from these particularities to find
+some nugget of valuable practices that have a more general application.
+Design patterns are a collection of reusable tried and tested solutions to recurring
+problems in engineering. They make our software more modular, maintainable, and extensible.
+Moreover, these patterns provide a common language for developers to use, making them
+an excellent tool for effective communication when problem-solving in teams.
```
pickfire, line 17: Can you please split this into 2 paragraphs? It seemed quite long to read.

intro.md
```diff
@@ -7,10 +7,14 @@ If you are interested in contributing to this book, check out the
...
 ## Design patterns
 
-When developing programs, we have to solve many problems.
-A program can be viewed as a solution to a problem.
-It can also be viewed as a collection of solutions to many problems.
-All of these solutions work together to solve a bigger problem.
+In software development, we often come across problems that share similarities
+regardless of the environment they appear in. Although the implementation details
+are crucial to solve the task at hand, we may abstract from these particularities to find
+some nugget of valuable practices that have a more general application.
```
dantho, line 13: ~~some nugget of valuable practices that have a more general application.~~
the common practices that are generically applicable.

intro.md
neithernut: > Also, I think the english words it uses is way harder compared to the original one, I think some people may have difficulties reading this.

Imo it should be ok as it's part of the introduction, which many readers will skip anyway. Clarity is not as important here as it is elsewhere. Can you point out a specific phrase that could be problematic to non-native speakers?

intro.md
MarkoM-dot: I like your amendment, thank you!

intro.md
MarkoM-dot: I agree with the point you make about separating the paragraph into two and that's about it.

PR #305, opened 2022-03-12
patterns: remove TODO item

MarcoIeni:
I don't think a code example is necessary there. But if we believe it is, we might want to create a GitHub issue instead of keeping the `TODO` in the book itself.

### DIFF
--- a/patterns/index.md
+++ b/patterns/index.md
@@ -19,7 +19,7 @@ whole classes of problems. Some of them are also patterns that are _unique_ to R
 ## YAGNI
 
 If you're not familiar with it, YAGNI is an acronym that stands for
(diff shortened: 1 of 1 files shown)


simonsan:
Could you open the issue before we merge this, so we don't forget it? ;)

MarcoIeni:
I just noticed that the [strategy](https://github.com/rust-unofficial/patterns/blob/main/patterns/behavioural/strategy.md) pattern is part of this book. So I don't think that adding an example here is necessary.

MarcoIeni:
Hey simonsan, shall we merge this? :)

simonsan (approved):

PR #309, opened 2022-05-26
update mdbook version to 0.4.18

MarcoIeni:


simonsan:
Sorry for late review, have been a bit inactive due to illness and holidays.

Issue #322, opened 2022-10-11
Rename RAII to OBRM


9SMTM6:
As the relevant section points out RAII isn't a great name for the concept. Rust actually has its own name for the concept, Ownership Based Resource Management, which I find vastly superior.

Of course we should keep references to RAII to make this easy to find and to ease adaption, but I'd like to switch to primarily OBRM terminology in that section. 

I'll probably get to doing that over the weekend, but anyone may feel free to do that themselfes. This issue was created to ensure this eventually gets resolved, and to ensure that the maintainers will accept such a PR.

9SMTM6:
Nomicon uses OBRM:

https://doc.rust-lang.org/nomicon/obrm.html

Relevant page in this project:

https://rust-unofficial.github.io/patterns/patterns/behavioural/RAII.html

neithernut:
Or at least add a short paragraph explaining that it's called RAII in other languages/in general. (including said link or paraphrasing the rationale).

9SMTM6:
There's different groups we could want to target with that page. As it it, that page is combining the bad parts of the target groups I can think of.

If we want to ease the understanding of programmers coming from C++, then yeah, the page quite openly uses that terminology, but these people already know patterns for RAII, and the page does NOT make a good job of explaining the subtle differences between Rust and C++, like destructive move, and how that might apply to patterns.

If we want to ease understanding of completely new people then the current terminology makes things hard to understand. EVEN  the linked page on cppreference predominantly uses Rust terminology (they implicitly explain ownership, also using terms like lifetime).

And OBRM is, as far as I understand it, the official Rust terminology.

As I said in the opening post, I want to keep RAII terminology present to address your point @simonsan, we could also keep it in the title, but if I an going to update the page I would use Rust terminology as a default. If that's not what the maintainers want then perhaps someone else should do it how it's desired.

neithernut:
> As I said in the opening post, I want to keep RAII terminology present to address your point @simonsan, we could also keep it in the title, but if I an going to update the page I would use Rust terminology as a default. If that's not what the maintainers want then perhaps someone else should do it how it's desired.

Imo RAII doesn't need to appear in the title at all (and, in fact, shouldn't if we switch to OBRM) as long as the page pops up in a search for "RAII". A short note that the concept is also known as/identical in its execution to RAII should be sufficient. However, it's necessary to have that small note in order to avoid confusion.

[...]

PR #323, opened 2022-10-16
Switch from RAII to OBRM Terms, expand and improve parts (#322)

9SMTM6:
This attempts to solve #322.

Since what was to be done finally was still up for discussion I've now mostly followed by opinion, and have not accounted for that of others.

This obviously isnt entirely finished and still needs polishing, but should serve as a basis for discussion.


9SMTM6:
markdownlint doesnt seem to have an autofix for the line-lenght fix.

As long as we're still discussing the fix I won't put in the work to manually fix these lints, housekeeping that result manually during changes is frankly no use of my time I accept.

9SMTM6:
Also, I should note that I dont have a TERRIBLE lot of experience with RAII in C++, so these sections are more at risk of being slightly wrong or not as complete as one could wish.

SUMMARY.md
```diff
@@ -26,7 +26,7 @@
     - [Command](./patterns/behavioural/command.md)
     - [Interpreter](./patterns/behavioural/interpreter.md)
     - [Newtype](./patterns/behavioural/newtype.md)
-    - [RAII Guards](./patterns/behavioural/RAII.md)
+    - [Resource management with OBRM](./patterns/behavioural/OBRM.md)
```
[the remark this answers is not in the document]
9SMTM6: Hmm you're right, there's also external links that have to be considered.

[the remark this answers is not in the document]
9SMTM6: I'll accept this change when I'm at a PC where I can also rename the file back, currently on mobile.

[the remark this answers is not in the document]
9SMTM6: In the meantime I've commited the change in a seperate commit, not sure why Github doesn't track that, but in that commit you can see a diff between before and after.

idioms/dtor-finally.md
```diff
@@ -1,5 +1,9 @@
 # Finalisation in destructors
 
+<!-- I'm not sure this is idomatic to Rust, usually one would want to handle that in the types used themself. -->
+<!-- doesnt draw comparisons to `defer` in eg golang or ziglang -->
+<!-- Is unneccessarily verbose, IF one would want to do that one could simply define a `Defer` type that holds a closure OR a function pointer that gets executed on drop -->
+<!-- theres also crates that aim to implenent this via macros, I expect they use something like the above or below. -->
```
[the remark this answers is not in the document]
9SMTM6: Yeah, I just noticed that while I was there, and didn't have time nor was sure how to approach that. But a seperate issue for that is totally fine by me.

patterns/behavioural/RAII.md
```diff
@@ -1,15 +1,27 @@
...
+* check unlinted formatting
+ -->
 ## Description
 
-[RAII][wikipedia] stands for "Resource Acquisition is Initialisation" which is a
-terrible name. The essence of the pattern is that resource initialisation is done
-in the constructor of an object and finalisation in the destructor. This pattern
-is extended in Rust by using an RAII object as a guard of some resource and relying
-on the type system to ensure that access is always mediated by the guard object.
+"Ownership Based Resource Management" (OBRM) - also known as ["Resource Acquisition is Initialisation" (RAII)][wikipedia] - is an idom meant to make handling resources easier and less error-prone.
```
wookietreiber, line 11: s/idom/idiom/

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
...
+
+* while C++ code often interfaces with C code or code in older styles, which doesn't use RAII, Rust was designed without the need to interface with such code, so its far less common to implement OBRM yourself
+* C++ doesn't have `Deref` nor a borrow checker, so code using RAII can not archive the same combination of safety and ergonomics
+* perhaps most importantly, Rust has different semantics when it comes to moving and copying of values, this will be expanded on below.
+
+C++ has complex rules for copying and moving of values, that Rust managed to simplify while keeping most advantages.
+In C++ behavior on a "move" (which is semantically meant to signify passing held resources to the moved-to value) is customizable in its move and move-assignment constructors.
+But after a variable has been "moved out of", it must still be accessable in C++.
+In Rust, a moved-out-of variable can not be used, only reassigned a new value (this is referred to as "destructive move"), and the behavior on a move is not customizable, instead a move simply copies the bytes of the moved-out value into the moved-into variable, and ensures the semantics of a destructive move.
```
neithernut, line 139: AFAIU you don't _reassign_ a value to a variable moved away. You _bind_ something new to a name which now happens to be unused. One consequence is that the new thing doesn't need to have the same type.

Also, I'd split this sentence into two, e.g.
```suggestion
In Rust, a moved-out-of variable can not be used (this is referred to as "destructive move"), though the name of that variable can be re-used.
The behavior on a move is not customizable, instead a move simply copies the bytes of the moved-out value into the moved-into variable, and ensures the semantics of a destructive move.
```

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
...
 
+When compared with RAII in C++, there are a few significant differences:
+
+* while C++ code often interfaces with C code or code in older styles, which doesn't use RAII, Rust was designed without the need to interface with such code, so its far less common to implement OBRM yourself
+* C++ doesn't have `Deref` nor a borrow checker, so code using RAII can not archive the same combination of safety and ergonomics
+* perhaps most importantly, Rust has different semantics when it comes to moving and copying of values, this will be expanded on below.
+
+C++ has complex rules for copying and moving of values, that Rust managed to simplify while keeping most advantages.
+In C++ behavior on a "move" (which is semantically meant to signify passing held resources to the moved-to value) is customizable in its move and move-assignment constructors.
+But after a variable has been "moved out of", it must still be accessable in C++.
```
neithernut, line 138: This paragraph is not exactly wrong, but an imo important thing to understand when working with C++ is that the rules _in the language_ are actually not _that_ incredibly complex when it comes to object lifetimes. But they are from another era, and the choices back then had some unfortunate consequences, some of which you detail on further down.
Also, the whole "move" in C++ is really more a convention rather than a concept that exists in the language. And so you end up with zombie objects which end up in an "undefined-but-recoverable" state (simplified for educational purposes) after their contents was moved to another instance.

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
...
+In Rust, a moved-out-of variable can not be used, only reassigned a new value (this is referred to as "destructive move"), and the behavior on a move is not customizable, instead a move simply copies the bytes of the moved-out value into the moved-into variable, and ensures the semantics of a destructive move.
+
+<!-- TODO this should be improved, I find it difficult to separate the creation and management of RAII Objects in the **-constructor - so at declaration time - from the one when using the RAII object. Feedbak welcome. -->
+This massively simplifies creation and management of OBRM Objects compared to C++, where one often has to do a lot more manual management of RAII classes - definition of the `destructor`, the `copy constructor`, the `copy assignment constructor`, the `move constructor` and the `move assignment constructor` all at once -, which is very error prone, and where RAII objects have to have a legal moved-out state, which often makes usage of these classes more problematic.
+For example, `unique_ptr`, the C++ equivalent to `Box`, can contain `nullptr`.
```
neithernut, line 143: ... and thus is more comparable to an `Option<Box>`, but less safe.

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
...
+
+Rust also moves values by default, which can be opted out by explicitly calling `Clone::clone` on each assignment, or on a Type level by implementing `Copy`.
+It is currently forbidden, and that is expected to continue, to implement `Copy` on a Type that implements `Drop` or contains a Type that implements `Drop`.
+This means that resource aquisition in Rust is a lot more explicit than in C++, as it can not happen during a simple assignment as it can in C++.
```
neithernut, line 147: You more or less captured already this already:
In C++, objects are expected to stay where they are (in memory) but you end up copying objects every now and then (and knowing when its acceptable is important for a C++ dev). Forbidding that is often an explicit choice and moving is also often more explicit.
In Rust, values tend to move around and if they are simple enough (like a Plain Old Datatype in C/C++), they can impl Copy. If they are not you have to impl Clone in order to multiply an instance, and cloning is always explicit (you have to call `Clone::clone`).

neithernut (commented):
Some comments from my side. I actually only have some minor nits about the content. Most of my ramblings are more motivated by @9SMTM6 expressing not being too familiar with C++.

patterns/behavioural/RAII.md
```diff
@@ -71,24 +83,31 @@ fn baz(x: Mutex<Foo>) {
...
-finalised. We can make use of the borrow checker in Rust to statically prevent
-errors stemming from using resources after finalisation takes place.
+OBRM is a useful pattern for ensuring resources are properly handled.
+The borrow checker in Rust will statically prevent
+errors stemming from using resources after the resource has been released.
 
 The core aim of the borrow checker is to ensure that references to data do not
-outlive that data. The RAII guard pattern works because the guard object
-contains a reference to the underlying resource and only exposes such
+outlive that data. The OBRM guard pattern works because the guard object
```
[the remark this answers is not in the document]
9SMTM6: It was some sort of attempt of not having to rewrite sections below, and also of introducing the guard terminology which is also used in rusts stdlib (eg with the mutex guard).

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
 using the guard object more ergonomic. Implementing a `get` method on the guard
 works just as well.
 
+When compared with RAII in C++, there are a few significant differences:
+
+* while C++ code often interfaces with C code or code in older styles, which doesn't use RAII, Rust was designed without the need to interface with such code, so its far less common to implement OBRM yourself
+* C++ doesn't have `Deref` nor a borrow checker, so code using RAII can not archive the same combination of safety and ergonomics
```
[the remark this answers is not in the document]
9SMTM6: > And then there's the whole issue of multi-threading...

could you elaborate/link some resource? Or are you simply referring to the general problem in C++, which presumably makes correct RAII harder too?

patterns/behavioural/RAII.md
9SMTM6: > AFAIU you don't reassign a value to a variable moved away. You bind something new to a name which now happens to be unused. One consequence is that the new thing doesn't need to have the same type.

I'm pretty sure you can and I have reused a variable that was moved out. In that case it has to have the same type, and it'll use the same memory, which is the purpose of that rule (and probably the reason C++ requires that an object still be usable after being moved out of), rust found a somewhat more elegant way.

If you want to reuse the name only yeah you can vcreate a new let binding shadowing the variable, but thats not what I'm referring to.

But perhaps thats a sign that should be made more explicit?

Otherwise splitting the section is a good suggestion.

patterns/behavioural/RAII.md
9SMTM6: You are correct, the rules are not really that complex. What is complex is executing it.

As you say, because moves are more of a convention in C++, which more or less mostly just added a move operation and let the developers implement it, you get undefined states for many objects even in the standard lib, with a few exceptions where behavior is defined in the standard, like unique_ptr AFAIK.

patterns/behavioural/RAII.md
9SMTM6: In terms of how its working under the hood if used correctly, yes. But many people work under the assumtion that unique_ptr will always or at least mostly not contain nullptr.

I'm not certain how to express that without overcomplications, and in the end, unique_ptr was introduced with similar aims to box, so I kept it simple at the expense of being perhapos not entirely correct.

patterns/behavioural/RAII.md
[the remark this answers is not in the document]
neithernut: Yes, I was simply referring to handling the problem in C++. And it's not just that you may need to keep re-entrant-safety in mind when working on some OBRM-implementation. Nothing stops you from placing a mutex _guard_/_lock_ into a `shared_ptr` and handing it out to multiple threads. Sure, it takes a certain kind of stupidity or ignorance to do so, but... well... humans do crazy stuff.
However, I wouldn't refer to that problem in this section because in Rust it's clearly irrelevant (safe for someone running around doing ridiculous `unsafe` stuff).

patterns/behavioural/RAII.md
neithernut: I think we're talking past each other. Are you referring to variables implementing `Copy`? Then I agree with you. But OBRM-Objects do, virtually by definition, implement `Drop` and thus not `Copy`.

I thought you meant creating a new binding using the same name. In this case you may end-up reusing the memory previously occupied by the moved value, but that being specified in the language would be news to me.

patterns/behavioural/RAII.md
neithernut: Not exactly sure but I think even for `unique_ptr` the standard doesn't mandate that the object moved from is in a defined state but only that a call to `std::unique_ptr::reset()` will transition it to a defined state. But I imagine any sane implementation will leave a moved-from `std::unique_ptr` in a reset-state, already.

patterns/behavioural/RAII.md
9SMTM6: We could fix the correctness though, by not saying unique_ptr is equivalent, but instead something like that it was introduced for the same purpose.

patterns/behavioural/RAII.md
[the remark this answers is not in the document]
9SMTM6: So we fix that section by, depending on how well it fits, either also referencing operator* etc, or by removing that claim -  and perhaps also just moving that part to a seperate point - , the remainder is fine as is?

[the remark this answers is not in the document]
neithernut: I'd just drop the bit about `Deref`. The rest is fine imo.
```suggestion
* C++ doesn't have a borrow checker, so code using RAII can not archive the same combination of safety and ergonomics
```

patterns/behavioural/RAII.md
neithernut: Yes, that would also be fine.

patterns/behavioural/RAII.md
```diff
@@ -108,6 +127,25 @@ Note that implementing `Deref` is not a core part of this pattern, it only makes
 using the guard object more ergonomic. Implementing a `get` method on the guard
 works just as well.
 
+When compared with RAII in C++, there are a few significant differences:
+
+* while C++ code often interfaces with C code or code in older styles, which doesn't use RAII, Rust was designed without the need to interface with such code, so its far less common to implement OBRM yourself
```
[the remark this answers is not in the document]
9SMTM6: I updated it. Perhaps you want to read it again whether that change is correct in your opinion.

patterns/behavioural/RAII.md
9SMTM6: Yes we probably are. Heres some code to ensure we understand each other now:

```rs
struct MoveSemantics {
    field: String,
}

struct WhateverElse(u32);

fn main() {
    let mut to_move_out_of = MoveSemantics {
        field: String::from("Whatever"),
    };
    // move value out of to_move_out_of,
    // afterwards its illegal to access
    let moved_to = to_move_out_of;
    // error
    // let attempt_access = to_move_out_of.field;
    // reassignment to variable is legal and uses the same memory 
    // (not for the string, which is a seperate allocation, but even if 
    // to_move_out_of would be on the heap it would be reused nonetheless AFAIK
    // this might be beneficial for some situations
    // needs to be the same type of course
    to_move_out_of = MoveSemantics {
        field: String::new(),
    };
    // shadowing of the variable name, doesnt use the same memory
    let to_move_out_of = WhateverElse(2);
}
```

I think I used the right terminology here, the `variable` is the name and refers to the memory, the `value` is whats written in the memory and abstractly also connected resources, and you can define a new variable which has the same name as a old one, which would be what you've described above as far as I understood it.

9SMTM6: I was reffering to the reassignment of the variable.

The reason I felt the need to mention that possibility is that its the one "usage" of the variable that is allowed, AFAIK all others are forbidden. So for completeness its required unless we find another formulation that covers only "usages" of a variable other than reassignment. Also some C++ programmers might be looking for something like that, and might otherwise wrongly think you cant do it.

We could say thats not worth the effort and strike it, or we expand on it, or we keep it as-is and hope people understand it correctly.

patterns/behavioural/RAII.md
9SMTM6: We are getting a bit sidetracked;-P.

I could rewrite that section to say something to the manner of "copy and move in C++ are just operations with connected conventions, which must be upheld across the respective copy (assignment) constructor and move (assignment) constructor and might also need adjustments to the destructor", which AFAIK would be the correct description but seems fairly complex right now, and more importantly isnt really how people usually think about copy vs move.

Other suggestions?

patterns/behavioural/RAII.md
9SMTM6: Rewrote it, but perhaps you can think of a more fluent way to express it. Otherwise feel free to resolve this.

patterns/behavioural/RAII.md
9SMTM6: I believe if we address the points above we might integrate this with the section above about (currently) "complex C++ rules", but as it is this section still holds new information as far as I can see.

I'd defer this until we've cleared up the parts above.

patterns/behavioural/RAII.md
neithernut: Just as I thought, I misinterpreted what you wrote. I think we're on the same page now.
The only case where you _can_ reuse a variable after passing it by value is if its type implements `Copy`, so maybe the following would work?
```suggestion
In Rust, a moved-out-of variable can not be used (this is referred to as "destructive move") unless its type implements [Copy].
The behavior on a move is not customizable, instead a move simply copies the bytes of the moved-out value into the moved-into variable, and ensures the semantics of a destructive move.
```

patterns/behavioural/RAII.md
neithernut: No, I think your change is perfectly fine.

patterns/behavioural/RAII.md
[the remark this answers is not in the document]
neithernut: LGTM

PR #326, opened 2022-11-27
New Section: Lenses and Prisms

jhwgh1968:
As the intro says, this is a pure functional concept that Rust doesn't often use directly, but can be used to help understand some very "functional" looking APIs, such as Serde.

It is a rather difficult concept, even for me, so the wording might not be the best. I'm open to any feedback and improvements.


jhwgh1968:
Hmm, it looks like I didn't finish some of tho code examples...

I'll be fixing the code in the next couple hours for hours. Hopefully that should improve clarity, too.

At the very least, what about the topic and the Serde idea? That's what motivated this.

jhwgh1968:
Sounds like a plan!

functional/lenses.md
```diff
@@ -0,0 +1,438 @@
+# Lenses and Prisms
+
+This is a pure functional concept that is not frequently used in Rust, because
+other idioms can achieve the same result and more advanced forms are not
+supported.
+Never the less, exploring the concept may be helpful to understand other
+patterns in Rust APIs, such as
+[visitors](../../patterns/behavioural/visitor.md) and "callbacks".
```
jhwgh1968, line 8: Note that I have copied links from other pages that seem to work in style. I don't know why CI is complaining this link is broken anyway.

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
+# Lenses and Prisms
+
+This is a pure functional concept that is not frequently used in Rust, because
+other idioms can achieve the same result and more advanced forms are not
+supported.
+Never the less, exploring the concept may be helpful to understand other
```
neithernut, line 6: I've seldomly seen `Nevertheless`  spelled as three words. Is there are reason why you did spell it like this?

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
...
+  [...]
+  "customer_id": 1048576332,
+}
+```
+
+Another one contains the account information:
+
+```json
+{ "primary_customer_id": 1048576332,
+  "accounts"; [
```
neithernut, line 34: Typo:
```suggestion
  "accounts": [
```

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
...
+that dynamic languages do.
+However, there are significant differences:
+
+1. Static type checking is maintained. Only types which "opt-in" to having a
+   customer ID (with an `impl CustomerId` block) can be placed into the
+   iterator or collection. Duck typed languages require this to be a runtime
+   check, which can then go wrong at runtime.
+1. The `impl CustomerId` block lives with the type. This means the type can
+   live in a different crate than definition of `CustomerId` itself,
+   maintaining extensibility.
```
neithernut, line 131: You may want to rephrase this a bit. Impls can also reside in the same module as the trait definition. _I_ get what you are saying, but people new to Rust may get confused a little by this statement. Maybe add something like `in this case` or `in this example`.

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
...
+/*
+trait RecordType {
+    fn view<T, L>(&self, lens: L) -> T
+        where L: FnOnce(&dyn RecordType) -> Option<T>
+    {
+        (lens)(self)
+    }
+}
+*/
+// all this is commented out because Rust does not support generics in traits
```
neithernut, line 191: Rust _does_ support generics in traits. You can declare generic functions in traits. And recently a subset of GATs made it into stable. You cannot use generics with _dynamic dispatch_. But you can even work around that one (ironically by using even more dynamic dispatching).

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
...
+
+#[derive(Deserialize)]
+struct IdRecord {
+    name: String,
+    customer_id: String,
+}
+```
+
+In C++, the equivalent would be to generate a very complex code template, with
+a lot of compile time logic -- if it's even possible.
```
neithernut, line 404: Yes, it is ;) But it may become virtually unreadable at some point and at least in the past there was a non-vanishing probability that you stumbled over differences between implementations (e.g. clang vs. gcc).

functional/lenses.md
```diff
@@ -0,0 +1,481 @@
...
+}
+```
+
+Here the composition is done through the generic `A`, which is defined as types
+with both traits.
+In order to literally compose lens functions, Rust uses `map` and `and_then`
+on `Option<T>` and iterator, which would allow each lens to be called
+sequentially on items.
+
+## Prisms: Lenses, but Sum Types
```
neithernut, line 301: Not sure whether the sum type part makes sense.

neithernut (commented):
I get the feeling that you beat around the bush a bit, but I guess that's ok if you consider the target audience.

However, I also get the feeling that you fail to highlight the actual core about what Lenses are. Not being as familiar with functional concepts as I'd like, I'll have to investigate this a bit myself (probably after work) and report back later.

For now, I just have some nit-pics.

functional/lenses.md
neithernut: Ok, after reading "Functional references: Lens and other Optics in Scala" you reference I now get how you ended up talking about sum-types. But I'd still choose a different term here.

Yes, if you talk about abstract typing the term sum-type is correct. However, it will probably be confusing here because when we talking about sum-types we usually talk about _vales_ being of a sum-type. That's true for enums and trait objects (and a reference to an object implementing some interface in Java), but the very example you choose (serde) is a prime example that prisms may still have value if you don't actually have any run-time variability at all in your final program (i.e. no sum-types are involved, at least not in that specific way).

I'd rather go for "Prisms: higher order Lenses" or something, even if it's also not really correct.

neithernut (changes requested):
Ok, I've read up on Lenses now and get why it felt something important was missing.

Lenses are all about accessing, and specifically getting and setting, members of some datatype. You missed to highlight this more explicitly. Imo that information should be right in the introduction: either in the very first paragraph or right after the heading `Basic Lenses: Like Rust Traits` (which now feels inadequate).

functional/lenses.md
jhwgh1968: True. In C++, anything is possible :smile:

jhwgh1968:
Thanks for your comments @neithernut. They are all good points.

I will re-write this a little to address them.

jhwgh1968:
I have done a fairly major re-write and reflowed and made the content more specific, tightening down my use of words as per commenst by @neithernut.

I will wait to fix the minor markdown issues until I can get a big picture review.

functional/lenses.md
```diff
@@ -0,0 +1,364 @@
...
+}
+```
+
+The version of `unique_ids` shown here allows any types to be in the iterator,
+so long as it has an attribute called `customer_id` which can be accessed by
+the function.
+This is how most functional languages operate on lenses.
+
+Rather than macros, they achieve this with a technique known as "currying".
+ That is, they "partially construct" the function, leaving the type of the
```
neithernut, line 156:
```suggestion
That is, they "partially construct" the function, leaving the type of the
```

functional/lenses.md
```diff
@@ -0,0 +1,364 @@
...
+
+This is a pure functional concept that is not frequently used in Rust.
+Never the less, exploring the concept may be helpful to understand other
+patterns in Rust APIs, such as [visitors](../patterns/behavioural/visitor.md).
+They also have niche use cases.
+
+## Lenses: Uniform Access Across Types
+
+A lens is a concept from functional languages that allows accessing parts of a
+data type in an abstract, unified way.[^1]
```
neithernut, line 11: "parts of a data type" sounds kind of strange, but I don't currently have a better idea how to express this either.

neithernut (commented):
I only had a quick look but it does appear a lot clearer now.

jhwgh1968:
Now that I've fixed the last set of linter issues, it's ready as far as I'm concerned!

neithernut:
The school-of-haskell link is still somehow misparsed as a local reference.

functional/lenses.md
```diff
@@ -0,0 +1,365 @@
...
+- [luminance](https://github.com/phaazon/luminance-rs) is a crate for drawing
+  computer graphics that uses lens API design, including proceducal macros to
+  create full prisms for buffers of different pixel types that remain generic
+- [An Article about Lenses in
+  Scala](https://web.archive.org/web/20221128185849/https://medium.com/zyseme-technology/functional-references-lens-and-other-optics-in-scala-e5f7e2fdafe)
+  that is very readable even without Scala expertise.
+- [Paper: Profunctor Optics: Modular Data
+  Accessors](https://web.archive.org/web/20220701102832/https://arxiv.org/ftp/arxiv/papers/1703/1703.10857.pdf)
+
+[^1]: [https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/a-little-lens-starter-tutorial](https://web.archive.org/web/20221128190041/https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/a-little-lens-starter-tutorial)
```
MarcoIeni, line 365:
```suggestion
[^1]: [School of Haskell: A Little Lens Starter Tutorial](https://web.archive.org/web/20221128190041/https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/a-little-lens-starter-tutorial)
```

Maybe this will fix the failing CI job

jhwgh1968:
It looks like that did indeed fix it. Thanks @simonsan

functional/lenses.md
```diff
@@ -0,0 +1,365 @@
...
+
+fn unique_ids_lens<T>(iter: impl Iterator<Item = T>) -> HashSet<u64>
+where
+    T: LensRef<Optics![customer_id], u64>, // any type with this field
+{
+    iter.map(|r| *r.view_ref(optics!(customer_id))).collect()
+}
+```
+
+The version of `unique_ids` shown here allows any types to be in the iterator,
```
MarcoIeni, line 150:
```suggestion
The version of `unique_ids` shown here allows any type to be in the iterator,
```

functional/lenses.md
```diff
@@ -0,0 +1,365 @@
...
+This highlights that `Deserializer` is similar to `CustomerId` from earlier,
+allowing any format parser which implements it to create `Value`s based on what
+it parsed.
+The `Value` trait is acting like a lens in functional languages.
+
+But unlike the `CustomerId` trait, the return types of `Visitor` methods are
+*generic*, and the concrete `Value` type is *determined by the Visitor itself*.
+
+Instead of acting as one lens, it effectively a family of
+lenses, one for each concrete type of `Visitor`.
```
MarcoIeni, line 248:
```suggestion
Instead of acting as one lens, it effectively acts as a family of
lenses, one for each concrete type of `Visitor`.
```
Adding the verb

jhwgh1968:
I have applied the minor changes @MarcoIeni requested, and it has now passed CI.

neithernut (approved):

PR #332, opened 2022-12-18
chore: update rust edition to 2021

MarcoIeni:


MarcoIeni:
> also check which idioms, patterns are still needed or useful. 

I think that nothing changes with our content if the reader has `2018` or `2021` as the `edition` field of their crate.
If you are suggesting reviewing all the book...this is going to take a lot of time, which I would love to have, but I don't :/

Instead, by merging this PR now, we make sure that the rust examples of this book can be compiled with rust edition 2021 👍 

Also, does it makes sense to release the `patterns 2021 edition` at the end of 2022? 😛

OT: it was around this period we started maintaining this project together. So happy "anniversary"! Thanks for everything you have done for this project and happy holidays 🎄

PR #336, opened 2023-01-04
New idiom: return moved arg on error

kriomant:


idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,60 @@
+# A succinct name for the pattern
+
+## Description
+
+If fallible method receives consumes argument, return it back inside error.
```
neithernut, line 5: You need to drop either `receives` or `consumes`. And maybe add a few `a`s and `an`s. Also, we're usually talking about `function`s rather than `methods`. Suggestion:

```suggestion
If a fallible function consumes an argument, return that argument back inside an error.
```

Another thing which you may want to highlight is that you usually do this for functions mutating or converting some operand.

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,60 @@
...
+        false
+    };
+    
+    println!("success: {}", success);
+}
+```
+
+## Motivation
+
+In case of error you may want to try some alternative way or to
```
neithernut, line 45:
```suggestion
In case of an error you may want to try some alternative way or to
```

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,60 @@
...
+retry action in case of non-deterministic function. But if argument
```
neithernut, line 46:
```suggestion
retry action in case of non-deterministic function. But if the argument
```

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,60 @@
...
+is always consumed, you are forced to clone it on every call, which
+is not very efficient.
+
+Standard library uses this approach in e.g. `String::from_utf8` method.
```
neithernut, line 50:
```suggestion
The standard library uses this approach in e.g. `String::from_utf8` method.
```

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,60 @@
...
+When given vector doesn't contain valid UTF-8, `FromUtf8Error` is returned.
```
neithernut, line 51:
```suggestion
When given a vector that doesn't contain valid UTF-8, a `FromUtf8Error` is returned.
```

neithernut (changes requested):
The rule makes sense imo, but the text could be improved.

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,62 @@
...
+## Description
+
+If a fallible function consumes an argument, return that argument back inside
+an error.
+
+## Example
+
+```rust
+pub fn send(value: String) -> Result<(), SendError> {
+    // Simulate non-deterministic fallible action.
```
MarcoIeni, line 12:
```suggestion
    println!("using {value} in a meaningful way");
    // Simulate non-deterministic fallible action.
```
We could write something like this to make it explicit that the function is using the value. Actually, it would be great if we could think of an example where the value is used for real.

idioms/return-moved-arg-on-error.md
```diff
@@ -0,0 +1,62 @@
...
+
+fn main() {
+    use std::mem::take;
+
+    let mut value = "imagine this is very long string".to_string();
+
+    let success = 'send: {
+        // Try to send value two times.
+        for _ in 0..2 {
+            value = match send(take(&mut value)) {
```
MarcoIeni, line 32: why do you use `take` here?
The code works even without it for me. 🤔

MarcoIeni (changes requested):

kriomant:
Any feedback?

MarcoIeni:
I have still [this](https://github.com/rust-unofficial/patterns/pull/336#discussion_r1070641216) unresolved comment

idioms/return-consumed-arg-on-error.md
```diff
@@ -0,0 +1,63 @@
+# Return consumed argument on error
+
+## Description
+
+If a fallible function consumes an argument, return that argument back inside
+an error.
```
dantho, line 6: Agree with the change to "consumed" instead of "moved".  But now I miss the Rust-specific nomenclature.  Can you tweak the description so it's not completely lost:

If a fallible function consumes **(moves)** an argument, return that argument back inside
an error.

kriomant:
> I have still [this](https://github.com/rust-unofficial/patterns/pull/336#discussion_r1070641216) unresolved comment

I put link to Playground there which shows that compilation fails without `take`.

MarcoIeni:
> > I have still [this](https://github.com/rust-unofficial/patterns/pull/336#discussion_r1070641216) unresolved comment
> 
> I put link to Playground there which shows that compilation fails without `take`.

I don't see the link.
[Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021) is the playground without the `take`.

kriomant:
Well, and it doesn't compile

```
Compiling playground v0.0.1 (/playground)
error[[E0308]](https://doc.rust-lang.org/stable/error-index.html#E0308): mismatched types
  --> src/main.rs:18:32
   |
18 |             value = match send(&mut value) {
   |                           ---- ^^^^^^^^^^ expected struct `String`, found `&mut String`
   |                           |
   |                           arguments to this function are incorrect
   |
note: function defined here
  --> src/main.rs:1:8
   |
1  | pub fn send(value: String) -> Result<(), SendError> {
   |        ^^^^ -------------
help: consider removing the borrow
   |
18 -             value = match send(&mut value) {
18 +             value = match send(value) {
   |
```

MarcoIeni:
The code from the playground I linked before compiles:
image

Can't you follow the compiler advice and remove `& mut`?

kriomant:
> Can't you follow the compiler advice and remove `& mut`?

You are right, it's my fault. Fixed example.

MarcoIeni (approved):

MarcoIeni:
Thanks Mikhail! I merged it 🎉
Simonsan, feel free to open a PR with your suggestions if you give this a read.
I merged the PR before your feedback because I think it's very complicated to work with long-lived PRs. I hope you don't mind :)

PR #339, opened 2023-01-18
Replace wording "abuse" with "misuse" in Deref anti-pattern

simonsan:
The wording is imho pretty clear connotated with sexual abuse and every time I read it I feel like it's misplaced. So I feel like replacing "abuse" with "misuse" is a better and clear wording.


MarcoIeni (approved):
Thanks!

PR #354, opened 2023-04-06
Add Sealed trait pattern

rbran:


patterns/behavioural/sealed.md
```diff
@@ -71,4 +74,6 @@ without compromising the compatibility with existing code uses of it.
 
 ## See also
 
-Blog post from [Predrag](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) about sealed, private and other pattern for traits.
+Blog post from
+[Predrag](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
```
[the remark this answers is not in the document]
pickfire: I read that article, I find it quite nice to read too. But do we have to put archive link?

neithernut (commented):
LGTM, but I feel it could be extended a bit more:

 * Sealed traits in rust are rather nieche, you don't usually use them. Well, unless you do. Imo this is one of those more "advanced" topics.
 * The example used (`OsStrExt`) may already be viewed as something of a special case, as it exists multiple types for different configurations with required methods that may or may not diverge from each other: `unix` and `WASI` are, on the surface, equivalent; `windows`, as always, differs. You do list those declarations, but imo that may just confuse some readers.
 * At the same time that very example also hints at an additional motivation for sealed traits: ensuring comparability not only between versions (if you are not capable or willing to come up with a stable interface) but also between multiple _configurations_.

patterns/behavioural/sealed.md
```diff
@@ -0,0 +1,79 @@
...
+sealed traits that need to be managed.
+
+## Discussion
+
+Sealed traits are a useful tool for creating a set of related behaviors that are
+intended to be used together without allowing other behaviors to be added from
+outside the crate.
+
+This restriction also allow future additions to the trait
+without compromising the compatibility with existing code uses of it.
```
pickfire, line 73: I think the original article is better as in it shows a table to how to limit certain scope which we don't mention here. It might be hard for users to discover those themselves.

patterns/behavioural/sealed.md
```diff
@@ -0,0 +1,79 @@
...
+implementations of this trait will need updated, what is assured by the
+sealed trait to only happen locally.
+
+## Example
+
+The standard library makes use of a sealed trait, one example is the
+`OsStrExt` trait for
+[unix](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html),
+[windows](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStrExt.html) and
+[wasi](https://doc.rust-lang.org/std/os/wasi/ffi/trait.OsStrExt.html).
```
pickfire, line 34: Can we have more examples, IIRC serde does this too.

rbran: I could not find a good example from serde. They all seem too complex and obscure.

patterns/behavioural/sealed.md
rbran: I'm afraid that it will be too complex, maybe is better to separate each level of sealing in a different pattern.

patterns/behavioural/sealed.md
rbran: I added a simpler example with a more defined objective.

Issue #310, opened 2022-06-19
Do you consider trait-to-extend as a anti-pattern


sudipghimire533:
# Trait-to-extend ( I don't know what else could I name it )
I am presenting my opinion for another anti-pattern. While I have treated this as anti-pattern some of my friend says it might be pattern instead. So depending upon the discussion we might add it to (anti-)pattern section

**Creating a new trait to extend external struct functionality**

## The pattern
Suppose this is the content of struct `Time` in any external crate:
```rust
enum Color {
    Red, Green, Blue, Yellow, Black, White, Gray, Custom(u8, u8, u8)
}

struct Style {
    background: Color,
    foreground: Color,
}

impl Style {
    pub fn new(fg: Color, bg: Color, ...other_properties) -> {
        Style {
           foreground: fg, background: bg,
        }
    }
}

```

And we are using this crate. Suppose now we start to use this `Style` struct. Now in this application we need to only create Dark style i.e `{fg: White. bg: Black}` and Light style i.e `{ fg: Black, bg: White }`. What I previoudly did and some of my colleges still do is something like:

```rust
use other_crate::{
    Style,
    Color::{Black, White},
};

pub trait LightAndDarK {
  fn get_light(..other_properties) -> Style;
  fn get_dark(..other_properties) -> Style;   
}

impl LightAndDark for Style {
   fn get_light(..other_properties) -> Style {
        Style::new(Black, White)
    }

    fn get_dark(..other_properties) -> Style {
        Style::new(White, Black)
    }
}

fn main() {
    let light_style = Style::get_light(..);
    let dark_style = Style::get_dark(..);
}

```

I Consider this as anti-pattern. What I prefer a better idea for such case is to:
```rust
use other_crate::{
    Style,
    Color::{Black, White},
};

pub struct DarkStyle;
impl DarkStyle {
    pub fn new(..other_prop) -> Style { Style::new(White, Black) }
}

pub struct LightStyle;
impl DarkStyle {
    pub fn new(..other_prop) -> Style { Style::new(Black, White) }
}

fn new() {
    let light_style = LightStyle::new();
    let dark_style = DarkStyle::new();
}
```

Here the main idea is: there is not a method we want in external struct, since we can't extend external item create a new local trait with desired method and implement it to external item.

What do you think of this? And also if maintainers are willing to add this (anti-)pattern, I would be glad to take responsibility to write this

neithernut:
[...]

PR #355, opened 2023-04-06
Add Inner Trait pattern

rbran:


burdges:
You'd avoid re-implementing the exposed methods if you require `InnerCar : Car` btw.

It's a useful pattern if you've very large crates or internally use `dyn InnerCar` lots.  It'll also become more useful whenever specialization arrives.

Yet, if your project has many small crates, then you'd typically achieve this using free `fn`s instead.
```
fn set_speed<C: Car>(car: &mut C, new_speed: f64) { ... }
fn get_acceleration<C: Car>(car: &mut C) -> f64 { ... }
```

patterns/behavioural/inner-trait.md
```diff
@@ -0,0 +1,105 @@
...
+//more Car implementations...
+```
+
+## Motivation
+
+This pattern allows developers to provide additional functionality to the
+implementation of a public trait without exposing that functionality as part of
+the public API. By using a private trait, developers can also improve the
+reusability of their code, since the private functionality can be reused across
+multiple implementations of the public trait.
```
neithernut, line 74: Suppose I'm using a trait `Car` from another crate `vehicles`, but for a very specific purpose requiring some additional code. Do you feel that a new trait `InnerCar` private to my own crate would fall under this pattern? If yes, this would be an alternative motivation, wouldn't it?

neithernut (commented):
I feel this is a specific manifestation of a more general pattern which may or may not warrant its own (sub)section including #310 and "type classes" (note the double quotes) as in

```rust
trait Foo {
    // ...
}

impl <T: ...> Foo for T {
    // ...
}
```

So maybe we want to broaden this into the "extension trait" pattern proposed in #310. If not we should draw a line bwtween those two at some point in the future (as in, if/when someone writes a section about "extension traits" or something similar).

TL;DR: LGTM, no objections against merging this as is.

patterns/behavioural/inner-trait.md
```diff
@@ -0,0 +1,105 @@
...
+In rust there is the public/private trait analogous, the private trait
+implementing the hidden functionalities, while the public trait exposes
+functionality.
+
+## See also
+
+Wikipedia [OOP Interface](https://en.wikipedia.org/wiki/Interface_%28object-oriented_programming%29).
+
+Blog post from [Predrag](https://web.archive.org/web/20230406211349/https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
+about sealed, private and other patterns for traits.
```
pickfire, line 105: I think once the other PR is merged, we need to link them to each others.

patterns/behavioural/inner-trait.md
```diff
@@ -0,0 +1,105 @@
...
+It is possible to define a private trait that implements all the
+methods of a public trait and also includes some private functions. This pattern
+can be used to provide additional functionality to the implementation of a
+public trait while keeping the private methods hidden from the public API.
+
+## Example
+
+This example demonstrate how a public trait `Car` can be implemented and include
+extra private methods using a auxiliary private trait `InnerCar`.
+
```
pickfire, line 14: I think we probably want to show a real example in standard library or in popular crates for this.

patterns/behavioural/inner-trait.md
pickfire: There is another layer on top of this is to prevent breaking changes, since we can now limit functionality without exposing public API.

patterns/behavioural/inner-trait.md
```diff
@@ -0,0 +1,105 @@
...
+```rust,ignore
+// trait that is public and part of the API
+pub trait Car {
+    fn get_speed(&self) -> f64;
```
pickfire, line 18: I wouldn't recommend using `get_` since `speed` will do, this looks javaish and make it unnecessary long for not much benefits, rust already have types for this, we don't have to spell it out in function name.

patterns/behavioural/inner-trait.md
```diff
@@ -0,0 +1,105 @@
...
+    fn accelerate(&mut self, duration: f64);
+    fn brake(&mut self, force: f64);
```
pickfire, line 20: I would recommend keeping this to at most 2 functions to make a minimal example.

rbran:
The more that I think about this pattern, the less I like it. It fells too complex/edge-case for a pattern, or maybe is a simple adaptation/equivalent for an "abstract-class" made by someone that is too stuck with the OOP mentality.

rbran:
I'll rethink this, if I got a more concise idea for a pattern, I'll create a new PR.

PR #365, opened 2023-04-18
Remove ref_slice example

BD103:
`ref_slice` is [deprecated](https://lib.rs/crates/ref_slice) and its functionality has been added to the [standard library](https://doc.rust-lang.org/stable/std/slice/fn.from_ref.html). This PR simply removes it as an example from the "Prefer Small Crates" page.

## Other considerations

Would it be useful to add a note redirecting to the `std` implementation, or should the example just be removed entirely?

---

I'm new to contributing here. I've read `CONTRIBUTING.md`, but I'm unable to run `markdownlint` on my current device. If you need me to do this or anything else, please ask!

### DIFF
--- a/src/patterns/structural/small-crates.md
+++ b/src/patterns/structural/small-crates.md
@@ -33,14 +33,14 @@ We should take advantage of this tooling, and use smaller, more fine-grained dep
 
 ## Examples
 
-The [`ref_slice`](https://crates.io/crates/ref_slice) crate provides functions
-for converting `&T` to `&[T]`.
-
 The [`url`](https://crates.io/crates/url) crate provides tools for working with
 URLs.
 
 The [`num_cpus`](https://crates.io/crates/num_cpus) crate provides a function to
 query the number of CPUs on a machine.
+
+The [`ref_slice`](https://crates.io/crates/ref_slice) crate provides functions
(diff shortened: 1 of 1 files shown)


neithernut:
Although the crate is now deprecated, it still makes for a good, useful example imo. But I do see the merit of at least adding a note marking it as a "historical" example. I don't thing a link to the function in `std` is necessary, though.

neithernut:
> What do you suggest I do instead? We could move the example to the bottom of the list and add `(historical)` next to it.

Yes, for example.

> If this were done, I'd either need to revert the previous commit or create a new PR. I don't know how much keeping the commit history clean is important for this project though.

I'm not a maintainer :sweat_smile:, but [`CONTRIBUTING.md`](https://github.com/rust-unofficial/patterns/blob/16a4e95d9982905483337a2939b3f383c3eb8bc6/CONTRIBUTING.md#important-note) says you shouldn't force-push (and start with a draft-PR). But instead of creating a new PR, you could revert your one commit and then mark the reference. The maintainers will then likely request that you sqash the revert (and then _do_ force-push) after its gone through review.

BD103:
Ok, this PR is ready for review. I'll see if I can get in contact with a maintainer to merge it.

MarcoIeni (approved):
Thanks!

PR #378, opened 2023-11-07
fix: use proper definite, and indefinite articles

pavpen:
…sure.md

* Minor phrasing updates to keep definite, and indefinite articles to proper grammar. (Good for pedantic readers who get distracted by improper grammar.)
* Not sure if there's a style guide on this point. I find it easier to follow an idea in properly flowing speech.


MarcoIeni (approved):
thanks!

PR #388, opened 2024-01-29
Add init-struct pattern

virtualritz:
See [here](https://xaeroxe.github.io/init-struct-pattern/) for a more exhaustive description.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -32,6 +32,7 @@
     - [Visitor](./patterns/behavioural/visitor.md)
   - [Creational](./patterns/creational/intro.md)
     - [Builder](./patterns/creational/builder.md)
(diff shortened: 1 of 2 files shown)


virtualritz:
> In general, I'm not sure, that this values an own article.
> 
> I also see it more as an idiom, than a pattern.

I find that many Rustaceans, even people using the language for a while are oblibious of this pattern/idiom. Especially people coming from other languages. So I though rather than adding it as a "footnote" somehere on an existing page, I make a dedicated one.

Today I ran into this again [here](https://github.com/infobip-community/infobip-api-rust-sdk/issues/13).
I would say overall I ran into people with considerable Rust experience being unaware of this maybe half a dozen times last year.

That's why I decided to write this. But it only took me 20mins. So if the PR is rejected, it's not a biggie. :grin:

PR #394, opened 2024-03-16
feat: render PDF version of book with `mdbook-pandoc`

### LINKED ISSUE #380: Add pdf renderer and link rendered book
We should render `patterns` also to a pdf-book. Following this discussion here, there is a nice pdf renderer coming: https://github.com/google/comprehensive-rust/issues/1543

max-heller:
Uses [mdbook-pandoc](https://github.com/max-heller/mdbook-pandoc) to render a PDF version of the book and publishes it to Github pages.

[Rendered](https://max-heller.com/patterns/patterns.pdf) ([job log](https://github.com/max-heller/patterns/actions/runs/8308838774/job/22739553404#step:5:12))

Related issue #380

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -31,6 +31,7 @@ runs:
     - name: Install mdbook-pandoc and related dependencies
       run: |
         cargo install mdbook-pandoc --locked --version 0.5.0
+        sudo apt-get update
         sudo apt-get install -y texlive texlive-luatex texlive-lang-cjk librsvg2-bin fonts-noto
         curl -LsSf https://github.com/jgm/pandoc/releases/download/3.1.12.2/pandoc-3.1.12.2-linux-amd64.tar.gz | tar zxf -
         echo "$PWD/pandoc-3.1.12.2/bin" >> $GITHUB_PATH
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -27,3 +27,11 @@ runs:
     - name: Install i18n-helpers
       run: cargo install mdbook-i18n-helpers --locked --version '${{ steps.mdbook-i18n-helpers-version.outputs.version }}'
       shell: bash
+
+    - name: Install mdbook-pandoc and related dependencies
+      run: |
+        cargo install mdbook-pandoc --locked --version 0.5.0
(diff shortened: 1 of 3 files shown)


simonsan:
Wow, thank you so much :) I changed your initial comment, so we don't close that issue, because I still need to do some stuff there, WRT to linking the pdf for example, so I'll close it later on. Just FYI. ;) 

Looked through it and it's ready to merge, when it's ready from your side. 🚀

simonsan (approved):

.github/workflows/install-mdbook/action.yml
```diff
@@ -27,3 +27,12 @@ runs:
     - name: Install i18n-helpers
       run: cargo install mdbook-i18n-helpers --locked --version '${{ steps.mdbook-i18n-helpers-version.outputs.version }}'
       shell: bash
+
+    - name: Install mdbook-pandoc and related dependencies
+      run: |
+        cargo install mdbook-pandoc --locked --version 0.5.0
```
MarcoIeni, line 33: can we use https://github.com/marketplace/actions/install-cargo-binstall so that we don't need to compile from scratch?

simonsan: I'll add that later!

simonsan: My idea is to merge it as a foundation so we are getting started with it. Then I will open a polishing PR, because I saw while looking through the pdf, that there are some issues with line width, when it comes to code blocks.

max-heller: > there are some issues with line width, when it comes to code blocks

line wrapping in code blocks is coming [very soon](https://github.com/max-heller/mdbook-pandoc/pull/72/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR17)!

simonsan:
Thanks a lot, again <3

.github/workflows/install-mdbook/action.yml
simonsan: I fixed it now by hand, but maybe I can just revert this, when it lands. Thanks for the update!

simonsan: Ah wait, I thought you linked to an issue :D if it's already there, it may not make sense at all to push the changes, and I may rather wait and see the outcome.

PR #396, opened 2024-03-16
update `mdbook-pandoc` to 0.6 and install with taiki-e/install-action

max-heller:
Adds line wrapping in code blocks

[Rendered](https://max-heller.com/patterns/patterns.pdf) ([log](https://github.com/max-heller/patterns/actions/runs/8309446293/job/22740801172))

Odd, I've tested locally and it works but it doesn't seem to be working in CI. Seems like the wrong version is running somehow.

Related: #395


simonsan:
Hey there, yes, I updated main with. The install-action was outdated, so the tools were not available. It should be fixed now. I'll close this, after rebasing and solving the merge conflict I saw, that there is only one line diff with a white space. 😬

PR #398, opened 2024-03-16
fix(PDF): install `texlive-latex-extra` to get `fvextra` to enable line wrapping in code blocks

max-heller:
[Rendered](https://max-heller.com/patterns/patterns.pdf)

Related to #395


simonsan (approved):
Thanks for the follow up! LGTM!

PR #399, opened 2024-03-17
feat(pdf): add pdf download

simonsan:


simonsan:
Link check failing is expected, as the PDF at this path will be only rendered, when this PR has been merged.

PR #410, opened 2024-07-18
Fix compilation error

koutheir:
Use expressions instead of statements when returning the value to the outer scope, and use `Option<NonNull<T>>` instead of raw pointers because it is safer and still ABI-compatible.


koutheir:
I noticed that there are other places where usage of raw pointers could be replaced by `NonNull<T>` or `Option<NonNull<T>>`.

simonsan:
@jhwgh1968 If you have any time in between, could you give it a review as well, please? Thank you! 🫂

simonsan (approved):

PR #414, opened 2024-08-11
fix: typo "is is" to "is" in src/idioms/ffi/accepting-strings.md

AtmolanderMimi:
very simple typo fix from:
`The example is is written to ensure that:`
into:
`The example is written to ensure that:`

This can be found in the first line under the Advantages category in `src/idioms/ffi/accepting-strings.md`.

*(now i can rest knowing this typo will not curse the readers with irreparable trauma ;) )*


MarcoIeni (approved):
Thanks!

PR #415, opened 2024-08-30
Update dtor-finally.md

AndreiTsN:
I'm not sure, but it looks like a typo in the function name. Now, it's a 'baz', but I think it should be 'bar'.

### DIFF
--- a/src/idioms/dtor-finally.md
+++ b/src/idioms/dtor-finally.md
@@ -23,7 +23,7 @@ fn bar() -> Result<(), ()> {
     // The dtor of _exit will run however the function `bar` is exited.
     let _exit = Foo;
     // Implicit return with `?` operator.
-    baz()?;
+    bar()?;
     // Normal return.
     Ok(())
 }


MarcoIeni:
I think here "baz" just means "run another function". Otherwise the recursion never ends.

AndreiTsN:
In fact, recursion ends up in a stack overflow (and the expected 'drop'=finally call won't happen and we get a good example of 'Disadvantages')

MarcoIeni:
> it'll be good to have more clarifications here if 'baz' is correct

How can we achieve this?

AndreiTsN:
I think, the simplest way:

fn baz() {
    // some code
}

fn bar() -> Result<(), ()> {
...
    baz()?;
    // Normal return.
    Ok(())
}

MarcoIeni:
this sounds like a great idea, pr welcome 🙌

AndreiTsN:
thanks a lot! A new pr added.

PR #419, opened 2024-10-03
Fix a typo and use rustc instead of Rustc

ShapelessCat:


simonsan:
Hey there, could you run `dprint fmt` on your PR, please?

You can install it via `cargo install --locked dprint`.

Thank you for your contribution!

simonsan (approved):

PR #426, opened 2025-01-19
Update compose-structs.md

detoxifiedplant:
change in indentation & styling for better reading


marcoieni (approved):
thanks!

marcoieni:
Can you find a solution to fix dprint?

E.g. we could use the `^^^` also for the other comment?

marcoieni:
Something like this:

```
    print_database(&db); 
    //              ^^ Immutable borrow of `db` happens here
    // *connection_string = "new string".to_string();
    // ^^^^^^^^^^^^^^^^^^ Mutable borrow is used here
```

detoxifiedplant:
@marcoieni sure, let me see what can be done to fix it

PR #427, opened 2025-01-20
fix: separate compiler error from code block

detoxifiedplant:
`dprint` forces the indentation of comments to be aligned with previous line's comment, alternate solution is to add new line between both lines,
current changes provides good readability.

### DIFF
--- a/src/patterns/structural/compose-structs.md
+++ b/src/patterns/structural/compose-structs.md
@@ -17,7 +17,7 @@ pattern often reveals smaller units of functionality.
 Here is a contrived example of where the borrow checker foils us in our plan to
 use a struct:
 
-```rust
+```rust,ignore
 struct Database {
     connection_string: String,
     timeout: u32,
@@ -44,6 +44,7 @@ fn main() {
 ```
 
 The compiler throws following errors:
+
 ```ignore
 let connection_string = &mut db.connection_string;
                         ------------------------- mutable borrow occurs here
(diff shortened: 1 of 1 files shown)


src/patterns/structural/compose-structs.md
```diff
@@ -38,12 +38,20 @@ fn main() {
...
-                         // here
+    print_database(&db);
+    *connection_string = "new string".to_string();
 }
 ```
 
+The compiler throws following errors:
+
+    let connection_string = &mut db.connection_string;
+    						------------------------- mutable borrow occurs here
```
marcoieni, line 49:
```suggestion
                            ------------------------- mutable borrow occurs here
```
replace tabs with spaces

src/patterns/structural/compose-structs.md
```diff
@@ -38,12 +38,20 @@ fn main() {
...
 
+The compiler throws following errors:
+
+    let connection_string = &mut db.connection_string;
+                            ------------------------- mutable borrow occurs here
+    print_database(&db);
+                   ^^^ immutable borrow occurs here
+    *connection_string = "new string".to_string();
+    ------------------ mutable borrow later used here
+
```
marcoieni, line 54: this code block needs to be of type text, otherwise the test runs it as if it is rust code

detoxifiedplant: @marcoieni should work now

detoxifiedplant:
@marcoieni should probably pass the tests now

marcoieni (approved):
thanks!

detoxifiedplant:
thank you for patience ✌️🏻

PR #428, opened 2025-02-24
Minor text improvements

proski:
Trying to clarify some parts of the book, whatever sounded unclear to me.

### DIFF
--- a/src/anti_patterns/borrow_clone.md
+++ b/src/anti_patterns/borrow_clone.md
@@ -36,13 +36,10 @@ between the two are not synchronized -- as if two completely separate variables
 exist.
 
 There are special cases -- `Rc<T>` is designed to handle clones intelligently.
-It internally manages exactly one copy of the data, and cloning it will only
-clone the reference.
-
-There is also `Arc<T>` which provides shared ownership of a value of type T that
-is allocated in the heap. Invoking `.clone()` on `Arc` produces a new `Arc`
-instance, which points to the same allocation on the heap as the source `Arc`,
-while increasing a reference count.
+It internally manages exactly one copy of the data. Invoking `.clone()` on `Rc`
+produces a new `Rc` instance, which points to the same data as the source `Rc`,
+while increasing a reference count. The same applies to `Arc`, the thread-safe
+counterpart of `Rc`.
 
 In general, clones should be deliberate, with full understanding of the
 consequences. If a clone is used to make a borrow checker error disappear,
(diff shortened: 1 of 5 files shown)


src/anti_patterns/borrow_clone.md
```diff
@@ -36,13 +36,10 @@ between the two are not synchronized -- as if two completely separate variables
...
-clone the reference.
-
-There is also `Arc<T>` which provides shared ownership of a value of type T that
-is allocated in the heap. Invoking `.clone()` on `Arc` produces a new `Arc`
-instance, which points to the same allocation on the heap as the source `Arc`,
-while increasing a reference count.
+It internally manages exactly one copy of the data. Invoking `.clone()` on `Rc`
+produces a new `Rc` instance, which points to the same data as the source `Rc`,
+while increasing a reference count. The same applies to `Arc`, the thread-safe
+counterpart of `Rc`.
```
proski, line 42: It sounded like the use of heap is specific to `Arc`. I believe the use of heap is irrelevant here. I combined the best parts of `Rc` and `Arc` paragraphs and mentioned `Arc` at the end.

src/anti_patterns/borrow_clone.md
```diff
@@ -63,11 +60,7 @@ should be understood fully before assessing whether the clone is required or
 not.
 
 Also be sure to always run `cargo clippy` in your project, which will detect
-some cases in which `.clone()` is not necessary, like
-[1](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone),
-[2](https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy),
-[3](https://rust-lang.github.io/rust-clippy/master/index.html#map_clone) or
-[4](https://rust-lang.github.io/rust-clippy/master/index.html#clone_double_ref).
+some cases in which `.clone()` is not necessary.
```
proski, line 63: The last link is broken. I can see old mentions of `clone_double_ref`, perhaps it was removed. More importantly, clippy is unlikely to recommend removing `.clone()` if the compiler requires it (which is the premise of this page). But clippy can find something else that could be improved, which would in turn make `.clone()` unnecessary (e.g. clippy might suggest passing arguments to a function by reference so the value is not consumed). No need to focus on clone-specific lints.

src/anti_patterns/deny-warnings.md
```diff
@@ -106,6 +106,6 @@ certain that there will be more deprecated APIs in the future.
   `rustc --help` for a general list of options
 - [rust-clippy] is a collection of lints for better Rust code
 
-[rust-clippy]: https://github.com/Manishearth/rust-clippy
+[rust-clippy]: https://github.com/rust-lang/rust-clippy
```
proski, line 109: The link was to a fork of the official rust-clippy repository. I don't see anything interesting in that fork. Let's link to the original.

src/functional/optics.md
```diff
@@ -275,7 +275,7 @@ struct TestStruct {
 generate_visitor!(TestStruct);
 ```
 
-Or do they?
+But let's actually try that approach.
```
proski, line 278: Another unclear "they".

src/patterns/ffi/wrappers.md
```diff
@@ -153,7 +153,7 @@ optimizations that C code cannot attain. Being denied access to certain
 shortcuts is the price Rust programmers need to pay.
 
 [^1]: For the C programmers out there scratching their heads, the iterator need
-not be read *during* this code cause the UB. The exclusivity rule also enables
+not be read *during* this code to cause the UB. The exclusivity rule also enables
```
proski, line 156: I hope I understood that correctly. It might be better to expand this text and moving it from the footnote.

marcoieni (approved):

marcoieni:
thanks!

PR #433, opened 2025-03-31
Update builder.md

nicsalv:
Cite advantage about code extension use-case and the fact that Rust does not support default values for function parameters.

### DIFF
--- a/src/patterns/creational/builder.md
+++ b/src/patterns/creational/builder.md
@@ -73,6 +73,9 @@ Prevents proliferation of constructors.
 
 Can be used for one-liner initialisation as well as more complex construction.
 
+When the target struct is updated with new fields, the builder can be updated
+to leave client code correct without changes.
+
 ## Disadvantages
 
 More complex than creating a struct object directly, or a simple constructor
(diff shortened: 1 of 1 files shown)


src/patterns/creational/builder.md
```diff
@@ -73,6 +73,9 @@ Prevents proliferation of constructors.
 
 Can be used for one-liner initialisation as well as more complex construction.
 
+When the target struct is updated with new fields, the builder can be updated
+to leave client code correct without changes.
```
marcoieni, line 77: I'm not sure what this means

src/patterns/creational/builder.md
```diff
@@ -81,9 +84,9 @@ function.
 ## Discussion
 
 This pattern is seen more frequently in Rust (and for simpler objects) than in
-many other languages because Rust lacks overloading. Since you can only have a
-single method with a given name, having multiple constructors is less nice in
-Rust than in C++, Java, or others.
+many other languages because Rust lacks overloading and default values for
+function parameters. Since you can only have a single method with a given name,
+having multiple constructors is less nice in Rust than in C++, Java, or others.
```
marcoieni, line 89: this change looks good 👍

src/patterns/creational/builder.md
nicsalv: Imagine you have a custom type (the target struct) that you want to extend with new behaviour. You would like to add new behaviour and use it in new code, without modifying existing legacies. Since the builder pattern encapsulates the construction logic, it's possible to update the builder so that new fields can be used in new code while safely ignored in older usages.

marcoieni: oh, ok, maybe we should rephrase "correct without changes" with "backwards compatible"?

Also, we haven't defined what the target struct is. Maybe we can find a better name for it.

marcoieni: Also, I would rephrase as the following to follow the "active form" instead of passive:

When you add new fields to the target struct, you can update the builder to leave client code backwards compatible.

PR #437, opened 2025-05-25
Add new pattern: use custom traits to avoid complex type bounds

neithernut:
Introducing traits as an abstraction should be obvious enough in many Rust projects. However, I often found myself introduce "random" type bounds involving `Fn` that would grow unwieldy first until I finally introduced a new trait. So maybe this is advise some people, especially new Rust users, will find useful.

I did not include a Motivation section because I felt it was redundant. I also did not (yet) write a Discussion section. There are probably some similar uses of type classes in functional programming languages that we could refer to.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -37,6 +37,7 @@
     - [Compose Structs](./patterns/structural/compose-structs.md)
     - [Prefer Small Crates](./patterns/structural/small-crates.md)
     - [Contain unsafety in small modules](./patterns/structural/unsafe-mods.md)
+    - [Use custom traits to avoid complex type bounds](./patterns/structural/trait-for-bounds.md)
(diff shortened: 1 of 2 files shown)


src/SUMMARY.md
```diff
@@ -37,6 +37,7 @@
     - [Compose Structs](./patterns/structural/compose-structs.md)
     - [Prefer Small Crates](./patterns/structural/small-crates.md)
     - [Contain unsafety in small modules](./patterns/structural/unsafe-mods.md)
+    - [Use custom traits to avoid complex type bounds](./patterns/structural/trait-for-bounds.md)
```
simonsan:
```suggestion
    - [Avoid complex type bounds with custom traits](./patterns/structural/trait-for-bounds.md)
```

PR #442, opened 2025-05-30
update(small-crates): added mitigation to disadvantages

dzmitry-lahoda:


src/patterns/structural/small-crates.md
```diff
@@ -26,10 +26,11 @@ fine-grained dependencies.
   has both versions 1.0 and 0.5. Since the `Url` from `url:1.0` and the `Url`
   from `url:0.5` are different types, an HTTP client that uses `url:0.5` would
   not accept `Url` values from a web scraper that uses `url:1.0`.
+  Use [workspace.dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) to neutralize.
```
marcoieni, line 29: This only helps in your crates and it's just a developer experience improvement, not in crates download from crates io. Right? 

Also it's not clear what "neutralize" means. 
Probably "helps in enforcing the same crate version across all your workspace members" or similar is more clear

dzmitry-lahoda: so not sure, may be need to rework whole section, because.
1. there is no problem of hell with one workspace and different compilation roots(if foo and bar package have different compilation trees)
2.  workspace deps allow to mitigate aligning version if possible
3. cargo patch allows to mitigate some compilation issues for versions.
4. even one compilation root can depend on 2 version of same lib, as long as there is no unified call stack for both.

So Rust is efficient in solving dependency hell, way more better of others.

image


image

dzmitry-lahoda: fixed, please take a look

src/patterns/structural/small-crates.md
```diff
@@ -26,10 +26,35 @@ fine-grained dependencies.
   has both versions 1.0 and 0.5. Since the `Url` from `url:1.0` and the `Url`
   from `url:0.5` are different types, an HTTP client that uses `url:0.5` would
   not accept `Url` values from a web scraper that uses `url:1.0`.
-- Packages on crates.io are not curated. A crate may be poorly written, have
+  Rust and Cargo provide several options to handle this disadvantage:
+  - [workspace.dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)
+    helps in enforcing the same crate version across all workspace members
+  - dependnecy can be renamed in cargo.toml and in Rust module import
```
marcoieni, line 32:
```suggestion
  - a dependency can be renamed in `Cargo.toml` and in Rust module import 
```
anyway, how renaming a dependency help with dependency hell?

src/patterns/structural/small-crates.md
```diff
@@ -26,10 +26,35 @@ fine-grained dependencies.
...
+    helps in enforcing the same crate version across all workspace members
+  - dependnecy can be renamed in cargo.toml and in Rust module import 
+  - `package.resolver= "2"` allows to have several indirect and direct dependncies to be in same workspace,
+     so sometimes it fails (for few reasons), one can:
+    - if resolver fails to find compatible versions,
+      can use [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section)
+      to pin dependecy accross `Cargo.lock`
+    - `resolver = "3"` added Rust-version awareness,
+       so incompatible crate versions (based on the `rust-version` field) are not selected,
+       further reducing potential dependency hell
```
marcoieni, line 40: isn't the latest resolver already the default of the last rust edition? If yes, why documenting this? People basically don't need to know I think because the default is already the best option, right?

src/patterns/structural/small-crates.md
```diff
@@ -26,10 +26,35 @@ fine-grained dependencies.
...
+   - [zepter](https://github.com/ggwpez/zepter) can be used to resolve feature resolution conflicts
+      
+- Packages on [crates.io](https://crates.io) are not curated. A crate may be poorly written, have
```
marcoieni, line 43:
```suggestion
- Some packages on [crates.io](https://crates.io) are not curated. A crate may be poorly written, have
```
let's avoid generalizing 😅

src/patterns/structural/small-crates.md
```diff
@@ -26,10 +26,35 @@ fine-grained dependencies.
...
   unhelpful documentation, or be outright malicious.
+  Rust ecosystem has next tools to mitigate these disadvantages:
```
marcoieni, line 45: what do you mean with "next tools"?

src/patterns/structural/small-crates.md
dzmitry-lahoda: why mentioning dependency hell if it does not exist?

PR #454, opened 2025-09-19
Update send_request function signature to return Result

RaoulLuque:
In the Motivation `send_request` has a return type, in the example it does not. This PR fixes this :)


simonsan:
Thanks! 🚀

PR #457, opened 2025-10-08
fix small spelling mistake in generics-type-classes

Bultjer:


orangecms (changes requested):
This is no spelling mistake:
"lest" means _to prevent something from happening_; in this instance, we refrain from making the module `pub`, preventing possible mistakes people could make outside our control.

Admittedly, it is a word that non-native speakers may not understand.
I think this is a good call to change the wording:

> Keep the module private to prevent outside users from inventing their own protocols.

simonsan:
Thanks! 🚀

PR #463, opened 2025-12-14
fix: resolve mdbook@0.5.0 compatibility errors

hwhsu1231:
Fixes errors in the publish.yml workflow run caused by upgrading mdbook to version 0.5.0 in the .env file:

https://github.com/rust-unofficial/patterns/actions/runs/20082242879/job/57611823327


simonsan:
Oh, wow, thank you, I was just checking this, actually. I wanted to fix it this Sunday and a few more things. Thank you for your PR and the effort. 🫂 (EDIT: I already made the changes on main in parallel. So I closed this.)

